PR #3, opened 2019-10-29
bundlecontext support

rotty3000:

### DIFF
--- a/org.osgi.test.common/src/main/java/org/osgi/test/common/CloseableBundleContext.java
+++ b/org.osgi.test.common/src/main/java/org/osgi/test/common/CloseableBundleContext.java
@@ -1,0 +1,274 @@
+package org.osgi.test.common;
+/*
+ * Copyright (c) OSGi Alliance (2019). All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import java.io.InputStream;
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+import java.util.Collections;
+import java.util.Dictionary;
+import java.util.IdentityHashMap;
+import java.util.Set;
+
+import org.osgi.framework.Bundle;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.BundleException;
+import org.osgi.framework.BundleListener;
+import org.osgi.framework.FrameworkListener;
+import org.osgi.framework.InvalidSyntaxException;
+import org.osgi.framework.ServiceFactory;
+import org.osgi.framework.ServiceListener;
+import org.osgi.framework.ServiceObjects;
+import org.osgi.framework.ServiceReference;
+import org.osgi.framework.ServiceRegistration;
+
+public class CloseableBundleContext implements AutoCloseable, InvocationHandler {
+
+	private final BundleContext					bundleContext;
+	private final Class<?>						host;
+	private final Set<ServiceRegistration<?>>	regs		= Collections
+		.synchronizedSet(Collections.newSetFromMap(new IdentityHashMap<>()));
+	private final Set<FrameworkListener>		fwListeners	= Collections
+		.synchronizedSet(Collections.newSetFromMap(new IdentityHashMap<>()));
+	private final Set<ServiceListener>			sListeners	= Collections
+		.synchronizedSet(Collections.newSetFromMap(new IdentityHashMap<>()));
+	private final Set<BundleListener>			bListeners	= Collections
+		.synchronizedSet(Collections.newSetFromMap(new IdentityHashMap<>()));
+	private final Set<Bundle>					bundles		= Collections
+		.synchronizedSet(Collections.newSetFromMap(new IdentityHashMap<>()));
+	private final Set<ServiceReference<?>>		services	= Collections
+		.synchronizedSet(Collections.newSetFromMap(new IdentityHashMap<>()));
+	private final Set<ServiceObjects<?>>		serviceobjects	= Collections
+		.synchronizedSet(Collections.newSetFromMap(new IdentityHashMap<>()));
+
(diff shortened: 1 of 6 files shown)


org.osgi.test.common/src/main/java/org/osgi/test/common/CloseableBundleContext.java
```diff
@@ -0,0 +1,239 @@
...
+import org.osgi.framework.Filter;
+import org.osgi.framework.FrameworkListener;
+import org.osgi.framework.InvalidSyntaxException;
+import org.osgi.framework.ServiceFactory;
+import org.osgi.framework.ServiceListener;
+import org.osgi.framework.ServiceObjects;
+import org.osgi.framework.ServiceReference;
+import org.osgi.framework.ServiceRegistration;
+
+public class CloseableBundleContext implements AutoCloseable, BundleContext {
```
bjhargrave: Implementing BundleContext is risky. We do not want to constrain what version of OSGi is used by the tester. By implementing BundleContext we place constraints here which concern me.

We need to think of another way to do this. Perhaps inject a dynamic proxy of BundleContext so we can hook into certain methods.

org.osgi.test.common/src/main/java/org/osgi/test/common/CloseableBundleContext.java
```diff
@@ -0,0 +1,239 @@
...
+
+	@Override
+	public <S> ServiceRegistration<S> registerService(Class<S> clazz, S service, Dictionary<String, ?> properties) {
+		ServiceRegistration<S> reg = bundleContext.registerService(clazz, service, properties);
+		regs.add(reg);
+		return reg;
+	}
+
+	@Override
+	public <S> ServiceRegistration<S> registerService(Class<S> clazz, ServiceFactory<S> factory,
```
bjhargrave: Not sure why we need to handle services since the framework will already handle cleaning when the bundle in stopped. Or is the point of doing this to deal with cleaning up between test method invocations?

org.osgi.test.common/src/main/java/org/osgi/test/common/CloseableBundleContext.java
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	@Override
+	public void removeBundleListener(BundleListener listener) {
+		bundleContext.removeBundleListener(listener);
+		listeners.remove(listener);
+	}
+
+	@Override
+	public void addFrameworkListener(FrameworkListener listener) {
```
bjhargrave: Not sure why we need to handle listeners since the framework will already handle cleaning when the bundle in stopped. Or is the point of doing this to deal with cleaning up between test method invocations?

org.osgi.test.junit4/src/main/java/org/osgi/test/junit4/BundleContextRule.java
```diff
@@ -0,0 +1,70 @@
...
+ * &#64;Test
+ * public void aTest() {
+ * 	Bundle bundle = rule.get()
+ * 		.getBundle();
+ * }
+ * </pre>
+ */
+public class BundleContextRule implements TestRule {
+
+	private CloseableBundleContext bundleContext;
```
bjhargrave: We should not use this type here. We can test in the finally if the context implements AutoClosable and then cast and call close.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/BundleContextExtension.java
```diff
@@ -0,0 +1,142 @@
...
+					makeAccessible(field).set(testInstance, getBundleContext(field.getType(), extensionContext));
+				} catch (Throwable t) {
+					ExceptionUtils.throwAsUncheckedException(t);
+				}
+			});
+	}
+
+	private static class CloseableResourceBundleContext implements CloseableResource {
+
+		private final CloseableBundleContext bundleContext;
```
bjhargrave: Same here. Lets not use this type directly.

org.osgi.test.common/src/main/java/org/osgi/test/common/CloseableBundleContext.java
```diff
@@ -0,0 +1,239 @@
...
+import org.osgi.framework.ServiceListener;
+import org.osgi.framework.ServiceObjects;
+import org.osgi.framework.ServiceReference;
+import org.osgi.framework.ServiceRegistration;
+
+public class CloseableBundleContext implements AutoCloseable, BundleContext {
+
+	private final BundleContext					bundleContext;
+	private final Set<ServiceRegistration<?>>	regs		= ConcurrentHashMap.newKeySet();
+	private final Set<Object>					listeners	= ConcurrentHashMap.newKeySet();
```
bjhargrave: listeners must be identity based and you must use separate collections per listener type (framework, bundle, service) since the same object can implement all the listener types yet only be registered for one type.

org.osgi.test.common/src/main/java/org/osgi/test/common/CloseableBundleContext.java
rotty3000: but testing is not bundles
all too often we have false positives because the variants of one test leak into other tests in the same class.

org.osgi.test.common/src/main/java/org/osgi/test/common/CloseableBundleContext.java
rotty3000: each test invocation.

org.osgi.test.common/src/main/java/org/osgi/test/common/CloseableBundleContext.java
rotty3000: used a dynamic proxy

org.osgi.test.junit4/src/main/java/org/osgi/test/junit4/BundleContextRule.java
rotty3000: done!

org.osgi.test.common/src/main/java/org/osgi/test/common/CloseableBundleContext.java
rotty3000: done!

org.osgi.test.common/src/main/java/org/osgi/test/common/CloseableBundleContext.java
```diff
@@ -0,0 +1,167 @@
...
+				} catch (BundleException e) {
+					// ignore?
+				}
+			});
+
+		regs.forEach(ServiceRegistration::unregister);
+		bListeners.forEach(bundleContext::removeBundleListener);
+		sListeners.forEach(bundleContext::removeServiceListener);
+		fwListeners.forEach(bundleContext::removeFrameworkListener);
+	}
```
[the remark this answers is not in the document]
rotty3000: true!

PR #4, opened 2019-10-29
filters util function

rotty3000:

### DIFF
--- a/org.osgi.test.common/pom.xml
+++ b/org.osgi.test.common/pom.xml
@@ -31,6 +31,13 @@
 		OSGi Test Common Support
 	</description>
 
+	<dependencies>
+		<dependency>
+			<groupId>org.osgi</groupId>
+			<artifactId>osgi.core</artifactId>
+		</dependency>
+	</dependencies>
+
 	<build>
 		<plugins>
 			<plugin>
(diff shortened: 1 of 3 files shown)


org.osgi.test.common/pom.xml
```diff
@@ -31,6 +31,13 @@
 		OSGi Test Common Support
 	</description>
 
+	<dependencies>
+		<dependency>
+			<groupId>org.osgi</groupId>
+			<artifactId>osgi.core</artifactId>
```
bjhargrave: If our intent is to make proper bundles of these jars, we need to make sure we use the lowest framework version possible to accomplish our goals. So using Core R7 means no Core R6 users can use our bundles.

org.osgi.test.common/src/main/java/org/osgi/test/common/Filters.java
```diff
@@ -0,0 +1,44 @@
...
+	 *
+	 * @param format a format string
+	 * @param args format arguments
+	 * @return filter
+	 */
+	public static Filter format(String format, Object... args) {
+		try {
+			return FrameworkUtil.createFilter(String.format(format, args));
+		} catch (InvalidSyntaxException ise) {
+			throw new AssertionError("Invalid filter", ise);
```
bjhargrave: Shouldn't this be one of the Junit assertion failure types? Or just duck the exception type and rethrow.

org.osgi.test.common/pom.xml
rotty3000: good point, reducing to 6

org.osgi.test.common/src/main/java/org/osgi/test/common/Filters.java
rotty3000: throwing ducks

PR #6, opened 2019-10-30
(wip) Junit5 test

rotty3000:
@bjhargrave suggested doing junit5 in a maven profile, but with this branch at least folks can try to see what needs to be done to get it working.


rotty3000:
pretty sure the issue is serviceloadery stuff with the test engine.

Also you need this small fix https://github.com/bndtools/bnd/issues/3509

rotty3000:
With the last change no need for bndtools/bnd#3509

Issue #23, opened 2019-11-14
[common] CloseableBundleContext to handle superclass methods


kriegfrj:
Was running some tests using the `BundleContextRule` and found that it blew up when I tried to call `toString()` on the proxied bundle context.

Looking at the code, it seems that `invoke()` will only handle method calls for methods declared in the `AutoCloseable` and `BundleContext` interfaces. For any other method call, it will throw `IllegalArgumentException`. This includes method calls for inherited methods, such as those inherited from `Object` and `BundleReference`. By a happy coincidence, the only method declared in `BundleReference` (`getBundle()`) is also declared in `BundleContext` (though without the `@Override` tag - should it have it?), so it still works. However, the methods of `Object` code would rightly expect to be available too.

For `CloseableBundleContext` to do the best possible job of masquerading as a true `BundleContext`, it needs to ensure that it handles all method calls.

Fix to come.

rotty3000:
The fix was merged.

PR #25, opened 2019-11-14
[common] CloseableBundleContext handle common Object methods

kriegfrj:
Looking at all of the primitive methods in `Object`, there are quite a few (in particular, those that deal with synchronization) that open up a can of worms with regards to how to handle them. So I have elected to override the ones that are most commonly used (`toString()` for example is often used in testing/debugging), and leave the `IllegalArgumentException` as the default behaviour for the others.

### DIFF
--- a/org.osgi.test.common/src/main/java/org/osgi/test/common/osgi/CloseableBundleContext.java
+++ b/org.osgi.test.common/src/main/java/org/osgi/test/common/osgi/CloseableBundleContext.java
@@ -84,6 +84,17 @@ public class CloseableBundleContext implements AutoCloseable, InvocationHandler 
 				return method.invoke(bundleContext, args);
 			}
 		}
+		if (method.getDeclaringClass()
+			.equals(Object.class)) {
+			switch (method.getName()) {
+				case "toString" :
+					return delegatedToString(proxy);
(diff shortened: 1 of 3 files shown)


kriegfrj:
Implements #23.

rotty3000:
before we move too far we I think we should figure out the flakiness of those tests.

kriegfrj:
I'm onto it - I think I've figured it out.

I think the issue is if a test is already registered before it starts listening for new tests. Now that we're calling `registerService()` with a delay of 0, the chances of this happening are high.

kriegfrj:
> I'm onto it - I think I've figured it out.
> 
> I think the issue is if a service is already registered before it starts listening for new tests. Now that we're calling `registerService()` with a delay of 0, the chances of this happening are high.

Hmm, I don't think that's it. Still looking...

kriegfrj:
> I think the issue is if a service is already registered before it starts listening for new tests. Now that we're calling `registerService()` with a delay of 0, the chances of this happening are high.

Hmm, I don't think that's it. Still looking...

kriegfrj:
> Hmm, I don't think that's it. Still looking...

Issue found & fixed (#28).

PR #31, opened 2019-11-15
[build] add baseline

rotty3000:


bjhargrave:
What are we baselining against? We have not even made the first release yet :-)

rotty3000:
this is just in preparation for that day!

PR #40, opened 2019-11-27
[junit5] ServiceUseExtension is not stateless

### LINKED ISSUE #37: [junit5] ServiceUseExtension is not stateless
The `ServiceUseExtension` keeps the state variables for the BundleContextExtension and the TrackServices instances.

State variables are not permitted in Jupiter extensions, because Jupiter makes no lifecycle guarantees about individual extension object instances (see https://junit.org/junit5/docs/current/user-guide/#extensions-keeping-state). Instead, extensions are supposed to store their state in the supplied `ExtensionContext` object.

rotty3000:
fixes #37

### DIFF
--- a/org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceUseExtension.java
+++ b/org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceUseExtension.java
@@ -25,11 +25,11 @@ import static org.osgi.test.common.filter.Filters.format;
 import java.lang.reflect.Field;
 import java.util.function.Predicate;
 
-import org.junit.jupiter.api.extension.AfterEachCallback;
 import org.junit.jupiter.api.extension.BeforeEachCallback;
 import org.junit.jupiter.api.extension.ExtensionConfigurationException;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
+import org.junit.jupiter.api.extension.ExtensionContext.Store.CloseableResource;
 import org.junit.jupiter.api.extension.ParameterContext;
 import org.junit.jupiter.api.extension.ParameterResolutionException;
 import org.junit.jupiter.api.extension.ParameterResolver;
@@ -41,7 +41,7 @@ import org.osgi.test.common.tracking.TrackServices;
 import org.osgi.test.junit5.context.BundleContextExtension;
 
 public class ServiceUseExtension<T> extends BaseServiceUse<T>
-	implements AfterEachCallback, BeforeEachCallback, ParameterResolver {
+	implements BeforeEachCallback, ParameterResolver {
 
 	public static class Builder<T> {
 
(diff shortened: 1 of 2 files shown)


rotty3000:
@kriegfrj we can't make it totally stateless, because we need to constructor args. We also need to be able to satisfy the contract for `BaseServiceUse.getTrackServices()`.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceUseExtension.java
```diff
@@ -52,8 +52,22 @@
 		private long			timeout		= TrackServices.DEFAULT_TIMEOUT;
 
 		/**
+		 * Create with default BundleContextExtension.
+		 *
 		 * @param serviceType of the service
 		 */
+		public Builder(Class<T> serviceType) {
+			this.serviceType = serviceType;
```
bjhargrave: Isn't this just `this(serviceType, new BundleContextExtension());` So it reuses the other constructor?

rotty3000: sure, fixing

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceUseExtension.java
```diff
@@ -108,33 +120,36 @@ public Builder(Class<T> serviceType, BundleContextExtension contextExtension) {
...
 		contextExtension.beforeEach(extensionContext);
-		trackServices.init(contextExtension.getBundleContext(extensionContext));
-		injectFields(extensionContext, extensionContext.getRequiredTestInstance(), ReflectionUtils::isNotStatic);
-	}
-
-	@Override
-	public void afterEach(ExtensionContext extensionContext) throws Exception {
-		close();
+		// set this in order to be able to honour the BaseServiceUse contract
+		trackServices = getTrackServices(extensionContext);
```
kriegfrj: Do we even need to honour the full `BaseServiceUse` contract? If we inject an instance of `ServiceAware<T>` instead (either as a parameter or as an instance field in the test), then the extension itself doesn't need to implement `ServiceAware<T>`. That's how they seem to have done it in `SoftAssertionsExtension`, and in fact all the extensions that I've seen - the old pattern with rules (where you access members of the rule directly) has gone by the wayside in favour of injecting state objects into the tests.

If that's too hard, maybe we can use it as-is and see how we go.

kriegfrj: I'm going to approve as-is because I think this is at the least an improvement, even if not the final solution.

kriegfrj:
> @kriegfrj we can't make it totally stateless, because we need to constructor args.

I don't think that constructor args count as state, because they are immutable.

> We also need to be able to satisfy the contract for `BaseServiceUse.getTrackServices()`.

See review comment below - do we even need to implement this part of the contract, if we are injecting a `ServiceAware<T>` instead?

PR #54, opened 2019-12-18
Install.bundle

rotty3000:


rotty3000:
@bjhargrave really just want to make sure the last commit here is good. the rest of cleanups.

rotty3000:
oops, I forgot to actually start the bundle :|

org.osgi.test.common/src/main/java/org/osgi/test/common/context/CloseableBundleContext.java
```diff
@@ -15,6 +14,10 @@
  * limitations under the License.
  */
 
+package org.osgi.test.common.context;
+
+import static aQute.lib.exceptions.Exceptions.duck;
```
bjhargrave: I think we should define our own duck method in commons rather than conditional packaging from Bnd. We don't know what conflicts may arise in the future.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/context/BundleContextExtension.java
```diff
@@ -58,6 +61,12 @@ public void afterEach(ExtensionContext extensionContext) throws Exception {
 		if (closeableResourceBundleContext != null) {
 			closeableResourceBundleContext.close();
 		}
+		bundleContextThreadLocal.remove();
+	}
+
+	@Override
+	public BundleContext getBundleContext() {
+		return bundleContextThreadLocal.get();
```
bjhargrave: This seems sketchy. How can you ensure this is called in the same thread as the setter? Or a thread that never set.

org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/context/BundleContextRuleTest.java
```diff
@@ -36,6 +39,37 @@
...
+	public void testInstallBundle_B() throws Exception {
+		Bundle bundle = null;
+
+		try (WithContextRule it = new WithContextRule(getClass())) {
+			bundle = it.rule.installBundle("foo/tbfoo.jar", false);
+
+			assertThat(bundle).extracting(Bundle::getState)
+				.is(new Condition<Integer>(state -> (state & INSTALLED) == INSTALLED, "Installed"));
+		} finally {
+			assertThat(bundle).extracting(Bundle::getState)
```
bjhargrave: I am not sure I understand the idea of asserting in the finally block. If the assert in the try block throws an exception, the assert in the finally block can mask it by throwing its own exception, no? I get you are trying to test the try-with-resources uninstalls the bundles. But the bundle may not have properly installed and you want the assert failure. Asserting the try-with-resources properly cleaned up should just happen after the try-with-resources statement rather than in its finally block so that the finally assert cannot shadow a failed assert in the try-with-resources block.

org.osgi.test.common/src/main/java/org/osgi/test/common/context/CloseableBundleContext.java
rotty3000: It's not just duck, it's all the other goodness in Exceptions that I envision using, *WithExceptions for instance, and `Exceptions.unchecked` etc...

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/context/BundleContextExtension.java
rotty3000: if you have a suggestion let's hear it!

org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/context/BundleContextRuleTest.java
rotty3000: Is this a style complain? Does it make a significant difference?

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/context/BundleContextExtension.java
rotty3000: to be clear the `beforeEach` method is called in the same thread. that's defined in the contract of the Extension.

rotty3000: https://junit.org/junit5/docs/current/api/org/junit/jupiter/api/extension/BeforeEachCallback.html

org.osgi.test.common/src/main/java/org/osgi/test/common/context/CloseableBundleContext.java
bjhargrave: So copy or shade everything we actually need. The jars we build are not necessarily used as OSGi bundles, so we should be careful to avoid our copy of the package conflicting with someone elses.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/context/BundleContextExtension.java
rotty3000: I cannot see how to interpret "before each test is invoked" other than to mean "before, in the same thread, each test is invoked"

org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/context/BundleContextRuleTest.java
bjhargrave: It is more than a style complaint. It is a complaint about shadowing assertion failures in the try block.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/context/BundleContextExtension.java
bjhargrave: Why does it have to be thread local? I thought we should not directly hold state? We put it in the extension context. A threadlocal is state. It is just state distributed across threads. And since getBundleContext() here is a public method, anyone can call it from any thread.

rotty3000: explain to me how I'm supposed to implement this method otherwise?

rotty3000: This stateless business doesn't make any real sense to be honest. I can see it being relevant only when using
```
@ExtendWith(FooExtension.class)
public class Tests {...}
```
Because there's certainly no other way to handle state.

However, the Junit5 documentation shows instance Extensions being used exactly like above:
https://junit.org/junit5/docs/current/user-guide/#extensions-registration-programmatic-instance-fields

bjhargrave: I don't know why you add the method and who is supposed to use it. But if you are going to store state, a threadlocal is not the place to do it since only one thread can see the state.

rotty3000: I changed it to not use a TL.

rotty3000: it's used to install bundles from the instance of the extension during a setup method or during the test method itself!

rotty3000: more specifically, it's used to implement the `InstallBundle` interface.

rotty3000:
@bjhargrave can you take a look again? I used injection instead.

Issue #56, opened 2019-12-19
junit4: Intermitted test failure in CI build ServiceUseRuleTest.successWhenServiceWithTimeout


bjhargrave:
I have seen this failure on several occasions in the CI build. Restarting the build can the succeed. So perhaps there is some timing error here.

```
[INFO] --- bnd-testing-maven-plugin:4.3.1:testing (testing) @ org.osgi.test.junit4 ---

[INFO] Matching glob *

[INFO] Bnd inferred -runee: JavaSE-1.8

required wires for org.eclipse.osgi_3.15.100.v20191114-1701 [0] []

required wires for org.apache.servicemix.bundles.junit_4.12.0.1 [1] []

required wires for org.assertj.core_3.13.2 [2] []

required wires for org.osgi.test.common_1.0.0.201912190658 [3] []

required wires for org.osgi.test.junit4-test_1.0.0.201912190658 [4] [osgi.wiring.host; filter:="(osgi.wiring.host=org.osgi.test.junit4)" -> osgi.wiring.host; bundle-version:Version="1.0.0.201912190658"; osgi.wiring.host="org.osgi.test.junit4"]

required wires for org.osgi.test.junit4_1.0.0.201912190658 [5] []

required wires for biz.aQute.tester_4.3.1.201911131708 [6] []

required wires for org.osgi.test.junit4-test_1.0.0.201912190658 [4] [osgi.wiring.host; filter:="(osgi.wiring.host=org.osgi.test.junit4)" -> osgi.wiring.host; bundle-version:Version="1.0.0.201912190658"; osgi.wiring.host="org.osgi.test.junit4"]

required wires for org.osgi.test.junit4.tb1_1.0.0.201912190658 [10] []

required wires for org.osgi.test.junit4.tb1_1.0.0.201912190658 [10] []

required wires for org.osgi.test.junit4.tb1_1.0.0.201912190658 [11] []

required wires for org.osgi.test.junit4.tb1_1.0.0.201912190658 [11] []

TEST successWhenServiceWithTimeout(org.osgi.test.junit4.service.ServiceUseRuleTest) <<< ERROR: expected:<null> but was:<[{org.osgi.test.junit4.types.Foo}={service.id=35, service.bundleid=5, service.scope=singleton, foo=bar}]>

org.junit.ComparisonFailure: expected:<null> but was:<[{org.osgi.test.junit4.types.Foo}={service.id=35, service.bundleid=5, service.scope=singleton, foo=bar}]>

[...]

PR #58, opened 2019-12-19
to better diagnose test failures always register services with marker…

rotty3000:
… property


rotty3000:
@bjhargrave this doesn't solve #56 but hopefully it will make diagnosis a little easier.

rotty3000:
This now fixes the issue. I've run 5 builds (1 local, 4 travis) without failure.

rotty3000:
Note, I had reproduced the issue locally!

PR #80, opened 2020-03-06
Build.cleanup

rotty3000:
This uses the new (in bnd 5.0.0) `bnd-process-tests` maven mojo to simplify integration test setup.

### DIFF
--- a/org.osgi.test.junit4/test.bndrun
+++ b/org.osgi.test.junit4/test.bndrun
@@ -30,10 +30,10 @@
 	ch.qos.logback.core,\
 	org.apache.felix.logback,\
 	slf4j.api
--runrequires: bnd.identity;id='${project.artifactId}-test'
+-runrequires: bnd.identity;id='${project.artifactId}-tests'
 -runbundles: \
 	assertj-core;version='[3.15.0,3.15.1)',\
 	org.apache.servicemix.bundles.junit;version='[4.12.0,4.12.1)',\
 	org.osgi.test.common;version='[1.0.0,1.0.1)',\
-	org.osgi.test.junit4-test;version='[1.0.0,1.0.1)',\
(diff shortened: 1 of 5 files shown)


org.osgi.test.junit4/test.bndrun
```diff
@@ -30,10 +30,4 @@
 	ch.qos.logback.core,\
 	org.apache.felix.logback,\
 	slf4j.api
--runrequires: bnd.identity;id='${project.artifactId}-test'
--runbundles: \
```
bjhargrave: I am not sure I like removing the `-runbundles` from the bndrun file. Is that necessary or a good idea?

org.osgi.test.junit5/test.bndrun
```diff
@@ -37,22 +37,6 @@
 	org.apache.felix.logback,\
 	slf4j.api
 -runrequires: \
-	bnd.identity;id='${project.artifactId}-test',\
+	bnd.identity;id='${project.artifactId}-tests',\
 	bnd.identity;id='${junit-jupiter-engine.bsn}',\
 	bnd.identity;id='${junit-platform-launcher.bsn}'
--runbundles: \
```
bjhargrave: I am not sure I like removing the `-runbundles` from the bndrun file. Is that necessary or a good idea?

org.osgi.test.junit4/test.bndrun
rotty3000: I mean it's recomputed each time already. So all this does is force us to have to commit changes to source anytime we make outside changes that affect the resolution result. All I personally care about for this is that tests actually run and pass :) we'd see a failure if the resolution was ever miscalculated due to human error.

... but, I'm not strongly arguing about it... just merely annoying for this scenario :)

rotty3000: re-added it...

PR #94, opened 2020-04-08
Add license-maven-plugin

kriegfrj:
Added license-maven-plugin to the parent pom and re-ran the build to make sure the update license header is added to all files.

### DIFF
--- a/org.osgi.test.assertj/src/main/java/org/osgi/test/assertj/bundle/AbstractBundleAssert.java
+++ b/org.osgi.test.assertj/src/main/java/org/osgi/test/assertj/bundle/AbstractBundleAssert.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) OSGi Alliance (2019). All Rights Reserved.
+ * Copyright (c) OSGi Alliance (2019-2020). All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.osgi.test.assertj.bundle;
 
 import static org.assertj.core.api.InstanceOfAssertFactories.LONG;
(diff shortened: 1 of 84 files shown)


kriegfrj:
I noticed that AssertJ was using this plugin and I thought it was neat. I thought this would be relatively uncontroversial but noticed that a couple of the files had 2018 as their start copyright year, which got obliterated by this change. Are they copy-and-paste errors or do we need to extend the inception year back to 2018 for select files?

bjhargrave:
This plugin seems to pretty blindly set all the included files to have the same copyright notice including specific copyright years. This is without regard to if the files actually changes in those years.

I have a script I wrote some time back which I have used to manage the copyright years based upon the actual commit information in git.

```script
git log --format="%H" --name-only --after="2019-12-31" | grep ".*\.java$" | sort | uniq | xargs -n 1 osgicopyright.sh
```

osgicopyright.sh
```script
#!/bin/sh

file=$1
if [ -n "$file" ]; then
	year=$2
	if [ -z "$year" ]; then
		year=$(git log -1 --date=short --format="%ad" -- "$file" | sed -E -e "s/([0-9]{4})\-.*/\1/")
	fi
else
	year=$(date "+%Y")
fi

sedscript="1,/\*\// {	# limit to opening comment
	/Copyright.*OSGi Alliance/ {	# limit to OSGi copyright statement
		# if ($year) or (xxxx, $year) then skip line
		/\(([0-9]{4}, )?$year\)/n
		# If (xxxx) or (xxxx, yyyy) then change to (xxxx, $year)
		s/\(([0-9]{4})(, [0-9]{4})?\)/\(\1, $year\)/
	}
}" 

if [ -n "$file" ]; then
	sed -E -e "$sedscript" -i "" $file
else
	sed -E -e "$sedscript"
fi
```

You need to specify the proper `--after` date year. The script will compute the year the file last changes and modify the copyright year notice to that year if necessary.

So the license plugin is a rather blunt instrument whereas my scripts are much precise. So I am not sure I care for the license plugin unless it can be configured to base the copyright years on actual git data.

kriegfrj:
> This plugin seems to pretty blindly set all the included files to have the same copyright notice including specific copyright years. This is without regard to if the files actually changes in those years.
>

Yes.
 
> I have a script I wrote some time back which I have used to manage the copyright years based upon the actual commit information in git.
> 
> So the license plugin is a rather blunt instrument whereas my scripts are much precise. So I am not sure I care for the license plugin unless it can be configured to base the copyright years on actual git data.

That seems fair. Any chance of porting your script into a Mojo? :smile:

bjhargrave:
> Any chance of porting your script into a Mojo?

I am not sure how possible that is given the script relies on unix commands like grep, sed, xargs, etc. as well as calling git. Seems like a lot of code writing to replace all that in a java mojo.

kriegfrj:
Yes, I thought it would be non-trivial... hence the :smile:

PR #96, opened 2020-04-09
Make BundleContextExtension usable without an instance

kriegfrj:
Generally speaking, Jupiter extensions should ~not~ be stateless - they should store their state in the engine's "Store". By using a well-known namespace, this means that they can be accessible to other extensions (like `ServiceUseExtension`).

The core of this PR is to make `BundleContextExtension.getBundleContext(ExecutionContext)` a static method. This is possible because the extension class doesn't store any state in its instance, but in the Store - so you don't actually need an instance of the class to fetch its state (like the proxied `BundleContext`). Being static, it can be called directly from (eg) `ServiceUseExtension` without needing a live `BundleContextExtension` instance.

Because `BundleContextExtension` also accesses the proxied `BundleContext` instance through the Store using the same method, it will get the same instance. The call to `getOrComputeIfAbsent()` takes care of any necessary synchronization or ordering issues - whoever calls it first transparently creates the instance, and whoever calls it subsequently for the same `ExecutionContext` will get the same instance.

### DIFF
--- a/org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/context/BundleContextExtension.java
+++ b/org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/context/BundleContextExtension.java
@@ -41,12 +41,11 @@ import org.osgi.framework.FrameworkUtil;
 import org.osgi.test.common.context.CloseableBundleContext;
 import org.osgi.test.common.install.InstallBundle;
 
-public class BundleContextExtension
-	implements AfterEachCallback, BeforeEachCallback, ParameterResolver {
+public class BundleContextExtension implements AfterEachCallback, BeforeEachCallback, ParameterResolver {
 
-	public static final String		BUNDLE_CONTEXT_KEY		= "bundle.context";
-	public static final String		INSTALL_BUNLDE_KEY		= "install.bundle";
-	public static final Namespace	NAMESPACE	= Namespace.create(BundleContextExtension.class);
+	public static final String		BUNDLE_CONTEXT_KEY	= "bundle.context";
+	public static final String		INSTALL_BUNDLE_KEY	= "install.bundle";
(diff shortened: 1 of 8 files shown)


bjhargrave:
> The call to `getOrComputeIfAbsent()` takes care of any necessary synchronization or ordering issues

How do you know this is the case? I don't see this in the javadoc.

kriegfrj:
> > The call to `getOrComputeIfAbsent()` takes care of any necessary synchronization or ordering issues
> 
> How do you know this is the case? I don't see this in the javadoc.

The store is backed by `org.junit.jupiter.engine.execution.ExtensionValuesStore`, which is in turn backed by a `ConcurrentMap`.

rotty3000:
@bjhargrave you ok with this? If so can we merge it so I can build off it in my quest to improve the service use extension?

bjhargrave:
> The store is backed by `org.junit.jupiter.engine.execution.ExtensionValuesStore`, which is in turn backed by a `ConcurrentMap`.

While this is good. It alone is not sufficient. But I looked at the code around the use of the ConcurrentMap and the MemoizingSupplier which defers the call to the function until after one of possibly multiple competing MemoizingSuppliers has been placed in the ConcurrentMap completes the solution.

> you ok with this?

Yes.

kriegfrj:
> > The store is backed by `org.junit.jupiter.engine.execution.ExtensionValuesStore`, which is in turn backed by a `ConcurrentMap`.
> 
> While this is good. It alone is not sufficient.

True, but it does show that it is at least *intended* for it to be thread safe... any lack of thread safety would be a bug and not simply because it wasn't designed to be thread safe.

> But I looked at the code around the use of the ConcurrentMap and the MemoizingSupplier which defers the call to the function until after one of possibly multiple competing MemoizingSuppliers has been placed in the ConcurrentMap completes the solution.
> 

:+1:

PR #98, opened 2020-04-09
Change BundleContextExtensionTest to use BundleAssert

kriegfrj:
As a bit of a test (and an application of the "eat your own produce" principle) I thought to update `BundleContextExtensionTest` to use `BundleAssert`. Seems to work nicely.

If you want to see what the diagnostics look like, try changing one of the expected states on one of the `isInState()` assertions to see what it looks like when the assertion fails.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -275,6 +275,12 @@
 				<scope>test</scope>
 			</dependency>
 			<dependency>
+				<groupId>org.osgi</groupId>
+				<artifactId>org.osgi.test.assertj</artifactId>
+				<version>${revision}</version>
+				<scope>test</scope>
+			</dependency>
+			<dependency>
 				<groupId>org.junit.jupiter</groupId>
 				<artifactId>junit-jupiter-api</artifactId>
 				<version>${junit-jupiter.version}</version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -274,6 +274,12 @@
 				<version>${assertj.version}</version>
 				<scope>test</scope>
 			</dependency>
+			<dependency>
+				<groupId>org.osgi</groupId>
+				<artifactId>org.osgi.test.assertj</artifactId>
```
bjhargrave, line 279: I don't think this is a good idea to put in the parent pom. It should only be in the project (like junit5) pom which can use the assertj project to test.

kriegfrj: But I'ved added it to dependencyManagement, not to the raw dependencies. Adding it to dependencyManagement simply sets the default scope and version for child projects *if* any of them decide to add the artifact to their own dependencies. It doesn't add it to the dependencies of all projects by default.

bjhargrave: My bad. The "local" view in the PR did not make it clear. I should have looked the the complete file.

PR #99, opened 2020-04-09
redesign ServiceUseExtension to make better use of annotations

rotty3000:


kriegfrj:
I like it! Forgive me if i missed it, but there don't appear to be any tests for cardinality > 1? And tests for sanity checking for cardinality < 0?

PR #100, opened 2020-04-10
add injection support to junit4

rotty3000:
.. and a whole lot more. I basically refactored the entire BundleContext and Service support to use injection in both JUnit 4 and 5. There are README's for both and hopefully you'll note a significantly improved usage.


rotty3000:
I did add many more tests including the missing negative tests for cardinality and timeout but also many more.

rotty3000:
.. Hopefully you don't find the FieldInjector too _borrowy_ (from junit5)!

org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/service/FilterArgumentsTest.java
```diff
@@ -0,0 +1,55 @@
...
+	public ServiceRule			sur	= new ServiceRule();
+
+	@InjectService(filter = "(%s=%s)", filterArguments = {
+		"objectClass", "org.osgi.service.log.LoggerFactory"
+	})
+	LogService					logService;
+	@InjectService(filter = "(%s=%s)", filterArguments = {
+		"objectClass", "org.osgi.service.log.LoggerFactory"
+	})
+	BaseServiceUse<LogService>	lsServiceAware;
```
bjhargrave, line 39: Shouldn't the field type be `ServiceAware<LogService>`?

org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/service/FilterArgumentsTest.java
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright (c) OSGi Alliance (2019). All Rights Reserved.
```
bjhargrave, line 2: All these new source files should just be `(2020)`.

org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/service/FilterTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+public class FilterTest {
+
+	@Rule
+	public ServiceRule			sur	= new ServiceRule();
+
+	@InjectService(filter = "(objectClass=org.osgi.service.log.LoggerFactory)")
+	LogService					logService;
+	@InjectService(filter = "(objectClass=org.osgi.service.log.LoggerFactory)")
+	BaseServiceUse<LogService>	lsServiceAware;
```
bjhargrave, line 35: Shouldn't the field type be `ServiceAware<LogService>`?

org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/service/MultiCardinalityServiceTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.osgi.framework.FrameworkUtil;
+import org.osgi.framework.ServiceRegistration;
+import org.osgi.test.common.annotation.InjectService;
+import org.osgi.test.common.dictionary.Dictionaries;
+import org.osgi.test.common.service.ServiceAware;
+import org.osgi.test.junit4.types.Foo;
+
+public class MultiCardinalityServiceTest {
+
+	private static List<ServiceRegistration<?>>	regisrations	= new CopyOnWriteArrayList<>();
```
bjhargrave, line 38: variable name spelling.

org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/service/OwnServiceTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class OwnServiceTest {
+
+	@Rule
+	public ServiceRule			sur	= new ServiceRule();
+
+	@InjectBundleContext
+	BundleContext				bundleContext;
+	@InjectService(cardinality = 0)
+	ServiceAware<LogService>	lsServiceAware;
```
bjhargrave, line 38: I would assume this test should be looking for `ServiceAware<Foo>` since it is named `OwnServiceTest` and the test registers a `Foo` service.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceConfiguration.java
```diff
@@ -54,7 +54,8 @@ public void close() throws Exception {
 
 	@Override
 	public String toString() {
-		return String.format("ServiceUseExtension [Class=\"%s\", filter=\"%s\", cardinality=%s, timeout=%s]",
+		return String.format(
+			"ServiceUseConfiguration [Class=\"%s\", filter=\"%s\", cardinality=%s, timeout=%s]",
```
bjhargrave, line 58: Name should be `ServiceConfiguration`.

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/FilterArgumentsTest.java
```diff
@@ -0,0 +1,60 @@
...
+public class FilterArgumentsTest {
+
+	@InjectService(filter = "(%s=%s)", filterArguments = {
+		"objectClass", "org.osgi.service.log.LoggerFactory"
+	})
+	LogService					logService;
+	@InjectService(filter = "(%s=%s)", filterArguments = {
+		"objectClass", "org.osgi.service.log.LoggerFactory"
+	})
+	BaseServiceUse<LogService>	lsServiceAware;
```
bjhargrave, line 37: `ServiceAware<LogService>`

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/FilterTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.osgi.test.common.annotation.InjectService;
+import org.osgi.test.common.service.BaseServiceUse;
+
+@ExtendWith(ServiceExtension.class)
+public class FilterTest {
+
+	@InjectService(filter = "(objectClass=org.osgi.service.log.LoggerFactory)")
+	LogService					logService;
+	@InjectService(filter = "(objectClass=org.osgi.service.log.LoggerFactory)")
+	BaseServiceUse<LogService>	lsServiceAware;
```
bjhargrave, line 33: `ServiceAware<LogService>`

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/OwnServiceTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.osgi.test.junit5.types.Foo;
+
+@ExtendWith(BundleContextExtension.class)
+@ExtendWith(ServiceExtension.class)
+public class OwnServiceTest {
+
+	@InjectBundleContext
+	BundleContext				bundleContext;
+	@InjectService(cardinality = 0)
+	ServiceAware<LogService>	lsServiceAware;
```
bjhargrave, line 38: `ServiceAware<Foo>`

org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/service/FilterArgumentsTest.java
rotty3000: So provided the type is at least ServiceAware you can use it. :) there are a couple of our higher order types that have some added methods. I'll document those in another pass. For instance, this type let's me introspect the `ServiceConfiguration` details which I'm using the assert function.

org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/service/FilterArgumentsTest.java
rotty3000: right right.. fixing

org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/service/FilterTest.java
rotty3000: same as above.

org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/service/OwnServiceTest.java
rotty3000: nice catch... :|

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceConfiguration.java
rotty3000: fixing

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/FilterArgumentsTest.java
rotty3000: see earlier comment.

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/FilterTest.java
rotty3000: see earlier comment.

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/OwnServiceTest.java
rotty3000: fixing

org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/service/FilterArgumentsTest.java
bjhargrave: Using BaseServiceUse is implementation specific. You just happen to know it will work. But if I implemented ServiceAware or subclassed ServiceAware and tried to use that, the injection code would see type is assignable but fail to set the field. I am not sure using is assignable in the injection code is wise since we cannot support any subtype of ServiceAware (or List).

rotty3000:
Updated!

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/FilterTest.java
rotty3000: fixed

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/FilterArgumentsTest.java
rotty3000: fixed

org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/service/FilterTest.java
rotty3000: fixed

org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/service/FilterArgumentsTest.java
rotty3000: fixed, now ckecks and expects exactly the types supported.

org.osgi.test.junit4/src/main/java/org/osgi/test/junit4/service/ServiceRule.java
```diff
@@ -0,0 +1,142 @@
...
+ * &#64;InjectService
+ * Foo foo;
+ *
+ * &#64;Test
+ * public void aTest() {
+ * 	// use foo
+ * }
+ * </pre>
+ */
+public class ServiceRule extends BundleContextRule implements AutoCloseable, MethodRule {
```
bjhargrave, line 55: It seems odd to me that ServiceRule extends BundleContextRule. If that is necessary, then perhaps we only need one JUnit 4 rule?

org.osgi.test.common/src/main/java/org/osgi/test/common/service/ServiceAware.java
```diff
@@ -21,9 +21,10 @@
 
 import org.osgi.annotation.versioning.ProviderType;
 import org.osgi.framework.ServiceReference;
+import org.osgi.test.common.tracking.TrackingConfig;
 
 @ProviderType
-public interface ServiceAware<T> {
+public interface ServiceAware<T> extends TrackingConfig {
```
bjhargrave, line 27: I know this extends helps you testing but it seems the wrong abstraction. For the testing can't we just inject a ServiceAware object and cast to BaseServiceUse or TrackingConfig since we know about our impl?

org.osgi.test.junit4/src/main/java/org/osgi/test/junit4/service/ServiceRule.java
kriegfrj: Unlike JUnit 5 Extensions which have the store,  there's no built-in mechanism in JUnit 4 rules to synchronize between rules. I'm guessing he's fine this to make the synchronization easier. But maybe composition would be better than inheritance?

org.osgi.test.common/src/main/java/org/osgi/test/common/service/ServiceAware.java
rotty3000: There was no reason to hide this interface from the end user. In fact it contains details that are quite useful for things like the assertj's diagnostic `.as(...)' function.

bjhargrave: > There was no reason to hide this interface from the end user. In fact it contains details that are quite useful for things like the assertj's diagnostic `.as(...)' function.

OK, Then why not just put those methods in ServiceAware and get rid of TrackingConfig? I don't see any separate need for TrackingConfig then.

rotty3000: because they are implemented at different layers.

bjhargrave: I don't see that. TrackServices implements it for no real reason. BaseServiceUse implements it (as well as ServiceAware) and delegates the methods to a TrackServices instance. So if you just move the methods from TrackingConfig to ServiceAware, you do not need TrackingConfig.

org.osgi.test.junit4/src/main/java/org/osgi/test/junit4/service/ServiceRule.java
rotty3000: I moved it back to composition but you'll note that 99% of the time the default constructor will be used, anyway, it's not a big issue.

org.osgi.test.common/src/main/java/org/osgi/test/common/service/ServiceAware.java
rotty3000: I pushed a change with cast instead.

rotty3000:
@bjhargrave I eliminated the TrackerConfig interface as suggested.

org.osgi.test.junit4/src/main/java/org/osgi/test/junit4/service/ServiceRule.java
```diff
@@ -0,0 +1,168 @@
...
+	@Override
+	public void close() throws Exception {
+		for (Iterator<Entry<InjectService, ServiceConfiguration<?>>> itr = configurations.entrySet()
+			.iterator(); itr.hasNext();) {
+			Entry<InjectService, ServiceConfiguration<?>> entry = itr.next();
+			entry.getValue()
+				.close();
+			itr.remove();
+		}
+		bundleContextRule.close();
```
bjhargrave, line 98: This may be a problem if the `bundleContextRule` object was supplied to the `ServiceRule(BundleContextRule)` constructor. That is, should our close result in closing a supplied BundleContextRule? If we created the BundleContextRule, as is done by the `ServiceRule()` constructor, then we obviously want to close it. 

Perhaps the simple answer is to get rid of the `ServiceRule(BundleContextRule)` constructor. We just always create a new BundleContextRule. I don't see any use of the `ServiceRule(BundleContextRule)` constructor in this PR.

rotty3000: that depends on if we consider the consequences of having:
```
@Rule BundleContextRule bcr = new BundleContextRule();
@Rule ServiceRule sr = new ServiceRule();
```
where now you have double injection happening.
as opposed to having:
```
@Rule BundleContextRule bcr = new BundleContextRule();
@Rule ServiceRule sr = new ServiceRule(bcr);
```
where the internal state of BundleContextRule is managed and avoids the double injection

bjhargrave: Fine. But then `sr.close()` must not close `bcr`.

bjhargrave: And what if JUnit calls `bcr.close()` before `sr.close()`? Will 'sr.close()` fail because of its dependence on a closed `bcr`?

rotty3000: did you look at the code of BCR close? it should be resonably implemented so it doesn't matter who calls it first?

rotty3000: it shouldn't matter if bcr is closed I don't think

rotty3000: ... this is why I had used inheritance...

bjhargrave: It looks like this will call ServiceTracker.close which will call BundleContext.ungetService for each tracked service. Assuming the real BundleContext is still valid, then no issue.

bjhargrave: >  this is why I had used inheritance...

That just means you always had your own copy of a BundleContextRule which is what I was saying by removing the ServiceRule(BundleContextRule) constructor. Same result with aggregation instead of inheritance. Inheritance is a problem since it limits future implementation options.

bjhargrave: A ServiceRule _is not_ a BundleContextRule. A ServiceRule just needs a BundleContext.

rotty3000:
more refactor

org.osgi.test.junit4/src/main/java/org/osgi/test/junit4/service/ServiceRule.java
rotty3000: fixed all this

PR #101, opened 2020-04-11
simplify several classes

rotty3000:


org.osgi.test.common/src/main/java/org/osgi/test/common/service/ServiceConfiguration.java
```diff
@@ -17,18 +17,28 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.SortedMap;
+import java.util.function.Function;
+
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.Filter;
+import org.osgi.framework.ServiceReference;
 import org.osgi.test.common.tracking.TrackServices;
```
bjhargrave, line 32: Is there still a good reason to leave TrackServices in its own package? At the minimum, we should move it into the common.services package. But perhaps we can go further and just merge it function directly into ServiceConfiguration?

rotty3000: Done!

org.osgi.test.common/src/main/java/org/osgi/test/common/service/ServiceConfiguration.java
```diff
@@ -125,36 +178,75 @@ public S getService(ServiceReference<S> reference) {
...
 
 	@Override
 	public S waitForService(long timeout) throws InterruptedException {
-		return getTracker().waitForService(timeout);
+		return tracker.waitForService(timeout);
 	}
 
-	private ServiceTracker<S, S> getTracker() {
-		return trackServices.tracker();
+	private class InnerCustomizer implements ServiceTrackerCustomizer<S, S> {
```
bjhargrave, line 209: Should this be a static inner class? I don't think it depends upon anything in the outer class.

rotty3000: fixed

PR #110, opened 2020-05-13
[assertj] AssertJ 3.16 updates

kriegfrj:
Number of updates that leverage new features of AssertJ 3.16:

* Assertion classes now use `failWithActualExpectedAndMessage()` where appropriate.
* No need to use the shim `asObjectAssert()` as `extracting()` is now public.
* Composable soft assertions using `SoftAssertionsProvider` interface.

### DIFF
--- a/org.osgi.test.assertj/src/main/java/org/osgi/test/assertj/promise/PromiseSoftAssertions.java
+++ b/org.osgi.test.assertj/src/main/java/org/osgi/test/assertj/promise/PromiseSoftAssertions.java
@@ -22,17 +22,4 @@ import org.osgi.util.promise.Promise;
 /**
  * Soft assertions for {@link Promise}s.
  */
-public class PromiseSoftAssertions extends SoftAssertions {
-	/**
-	 * Create a soft assertion for a {@link Promise}.
-	 *
-	 * @param actual The {@link Promise}.
-	 * @param <T> The type of the value contained in the {@link Promise}.
-	 * @return The created soft assertion.
-	 */
-	public <T> PromiseAssert<T> assertThat(Promise<? extends T> actual) {
-		@SuppressWarnings("unchecked")
-		PromiseAssert<T> softly = proxy(PromiseAssert.class, Promise.class, actual);
-		return softly;
-	}
-}
+public class PromiseSoftAssertions extends SoftAssertions implements PromiseSoftAssertionsProvider {}
--- a/org.osgi.test.assertj/src/main/java/org/osgi/test/assertj/bundle/AbstractBundleAssert.java
+++ b/org.osgi.test.assertj/src/main/java/org/osgi/test/assertj/bundle/AbstractBundleAssert.java
@@ -106,7 +106,8 @@ public abstract class AbstractBundleAssert<SELF extends AbstractBundleAssert<SEL
 	public SELF hasSymbolicName(String expected) {
 		isNotNull();
 		if (!Objects.equals(actual.getSymbolicName(), expected)) {
-			failWithMessage("%nExpecting%n  <%s>%nto have symbolic name:%n  <%s>%n but was:%n  <%s>", actual, expected,
+			failWithActualExpectedAndMessage(actual.getSymbolicName(), expected,
+				"%nExpecting%n  <%s>%nto have symbolic name:%n  <%s>%n but was:%n  <%s>", actual, expected,
 				actual.getSymbolicName());
 		}
 		return myself;
@@ -121,7 +122,8 @@ public abstract class AbstractBundleAssert<SELF extends AbstractBundleAssert<SEL
 	public SELF hasLocation(String expected) {
 		isNotNull();
 		if (!Objects.equals(actual.getLocation(), expected)) {
-			failWithMessage("%nExpecting%n  <%s>%nto have location:%n  <%s>%n but was:%n  <%s>", actual, expected,
+			failWithActualExpectedAndMessage(actual.getLocation(), expected,
+				"%nExpecting%n  <%s>%nto have location:%n  <%s>%n but was:%n  <%s>", actual, expected,
 				actual.getLocation());
 		}
 		return myself;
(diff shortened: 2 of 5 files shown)


org.osgi.test.assertj/src/main/java/org/osgi/test/assertj/promise/PromiseSoftAssertions.java
```diff
@@ -22,17 +22,4 @@
...
-	 * @param <T> The type of the value contained in the {@link Promise}.
-	 * @return The created soft assertion.
-	 */
-	public <T> PromiseAssert<T> assertThat(Promise<? extends T> actual) {
-		@SuppressWarnings("unchecked")
-		PromiseAssert<T> softly = proxy(PromiseAssert.class, Promise.class, actual);
-		return softly;
-	}
-}
+public class PromiseSoftAssertions extends SoftAssertions implements PromiseSoftAssertionsProvider {}
```
bjhargrave, line 25: Should we not get rid of this class and just have the interface?

kriegfrj: We certainly can. I kept it for backward compatibility in case anyone was using it (including the test cases). But I suppose we're in pre-release and anyone using it should be ready to handle such incompatible changes.

kriegfrj: I also think an uber-interface like `OSGiSoftAssertionsProvider` would be a be a good move as we get more of these.

bjhargrave: I am not super excited by uber types. They are non-cohesive. I think it is better that each package be focused and we don't end up with a dumping ground like java.util.

bjhargrave: We should keep the `PromiseSoftAssertions` class. I just used it :-)

kriegfrj: > I am not super excited by uber types. They are non-cohesive. I think it is better that each package be focused and we don't end up with a dumping ground like java.util.

But the uber-types for the assertion entry points are one of the main features of AssertJ! They allow you to type `softly.assertThat().` without having to make sure you've instantiated just the exact correct `SoftAssertionsProvider` for the object that you're asserting on - it's already there. And when you're writing real-world tests (and especially using soft assertions), it is usual that you will want to assert on objects of multiple different types, and that you will want these assertion failures (if any) to be reported in a group in the order that they are executed.

AssertJ itself does not follow the pattern of defining one `SoftAssertionsProvider` per assertion type.

bjhargrave: > AssertJ itself does not follow the pattern of defining one `SoftAssertionsProvider` per assertion type.

But AssertJ is the the base and a complete entity. Like Java itself. We are just an appendage. From a package versioning point of view, such non-cohesive types create lots of changes to the uber package as each piece changes.

With the provider, it is now super easy for any user of our custom assertions and someone else's custom assertions to easy make what ever uber type makes sense for them, no?

kriegfrj: > > AssertJ itself does not follow the pattern of defining one `SoftAssertionsProvider` per assertion type.
> 
> But AssertJ is the the base and a complete entity. Like Java itself. We are just an appendage.

We are an appendage, but a complete one. It is natural for OSGi assertion types to be bundled together - literally! (we're not creating an individual bundle for each assertion type).

> From a package versioning point of view, such non-cohesive types create lots of changes to the uber package as each piece changes.
>

I don't see how an uber-class would be a source of too many versioning changes - it only would be a MINOR change when you add a new assertion type (which is infrequent), or a MAJOR change when you remove one (even less frequent).

With the one-per-assertion model, you end up with a lot more soft assertion provider classes that are a potential source of versioning changes to the packages that they live in.

> With the provider, it is now super easy for any user of our custom assertions and someone else's custom assertions to easy make what ever uber type makes sense for them, no?

Well, I suppose it's not too bad - but it's even easier to add one for all of osgi-test than it is to try and add VersionSoftAssertionProvider, BundleSoftAssertionProvider, PromiseSoftAssertionProvider, etc individually. It's easy now because we've only got a few of them, but what about when we've got dozens? Are we going to require every serious user of osgi-test to manually build their own uber-soft assertions entry point in this way?

rotty3000:
If your test used more than one or two soft assertion types your test is
probably too big, no?

org.osgi.test.assertj/src/main/java/org/osgi/test/assertj/promise/PromiseSoftAssertions.java
bjhargrave: > Are we going to require every serious user of osgi-test to manually build their own uber-soft assertions entry point in this way?

OK, lets try to make our uber type. Where does it live? In the base `org.osgi.test.assertj` package? Or in a new subpackage called ??

bjhargrave:
> If your test used more than one or two soft assertion types your test is probably too big, no?

I can see a test case need bundle and version asserts and then maybe filter or promise asserts depending upon the API of the bundle under test. It would be rare to use all, but probably likely to need more than one.

kriegfrj:
> > If your test used more than one or two soft assertion types your test is probably too big, no?
> 
> I can see a test case need bundle and version asserts and then maybe filter or promise asserts depending upon the API of the bundle under test. It would be rare to use all, but probably likely to need more than one.

And also @ray, you're right that a single test that uses many soft assertion types is probably too big, but across your suite of tests it's reasonable (even common) to use several.

I have created another issue to discuss this as I thought it was important to capture the reasoning for posterity: #115

PR #120, opened 2020-05-27
[assertj] Added ServiceEventAssert and FrameworkEventAssert

kriegfrj:


org.osgi.test.assertj/src/main/java/org/osgi/test/assertj/event/AbstractBitmappedTypeEventAssert.java
```diff
@@ -0,0 +1,93 @@
...
+		return getType.applyAsInt(actual);
+	}
+
+	public SELF isOfType(int expected) {
+		isNotNull();
+		if ((expected & (expected - 1)) != 0) {
+			throw new IllegalArgumentException(
+				"Multiple bits set in expected (" + expected + ") - do you mean to use isOfTypeMaskedBy()?");
+		}
+		final String expectedString = bitmap.toString(expected);
```
bjhargrave, line 47: We don't need the expectedString unless there is a failure, right? So move it inside the if.

org.osgi.test.assertj/src/main/java/org/osgi/test/assertj/event/AbstractBitmappedTypeEventAssert.java
```diff
@@ -0,0 +1,93 @@
...
+		if ((actualType() & expected) == 0) {
+			failWithMessage("%nExpecting%n <%s>%nto be of type:%n <%d:%s>%n but was of type:%n <%s>", actual, expected,
+				expectedString, bitmap.maskToString(actualType()));
+		}
+		return myself;
+	}
+
+	public SELF isNotOfType(int expected) {
+		isNotNull();
+		final String expectedType = bitmap.toString(expected);
```
bjhargrave, line 57: We don't need the expectedString unless there is a failure, right? So move it inside the if.

org.osgi.test.assertj/src/main/java/org/osgi/test/assertj/event/AbstractBitmappedTypeEventAssert.java
kriegfrj: Done.

org.osgi.test.assertj/src/main/java/org/osgi/test/assertj/event/AbstractBitmappedTypeEventAssert.java
kriegfrj: Done.

kriegfrj:
Thanks for the review!

PR #122, opened 2020-05-29
[assertj] Export events package that forms part of public API through inheritance

kriegfrj:
Fixes an issue pointed out by @bjhargrave:

Because `org.osgi.test.assertj.(service|bundle|framework)event` contain classes that have `org.osgi.test.assertj.event.AbstractBitmappedTypeEventAssert` as their superclass, and because a client bundle won't be able to resolve these classes unless they can also resolve the superclass, it is necessary to export `org.osgi.test.assertj.event` as well. Otherwise the client bundles (which will declare the `event` subpackage in their `Import-Package` declaration) will not be able to resolve.

Bnd warned about this during the build, both in Bndtools and in the Maven console, but because it was a warning it went unnoticed. I have queried @pkriens as to if there is a good reason why this was only a warning, as I think a strong case could be made for promoting the issue to a build error.

### DIFF
--- a/org.osgi.test.assertj/src/main/java/org/osgi/test/assertj/event/package-info.java
+++ b/org.osgi.test.assertj/src/main/java/org/osgi/test/assertj/event/package-info.java
@@ -1,0 +1,19 @@
+/*
+ * Copyright (c) OSGi Alliance (2020). All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


rotty3000:
Was the warning about exporting a package that contained references to private packages?

The reason, I believe, this an warning rather than an error is because the moment you start to make all OSGi _issues_ errors nothing in the wild will ever build. This specific issue is a delicate one, particularly when migrating non-OSGi code into OSGi.

Now, if you want to hoist this particular warning in your project to an error, bnd can do that with:

`-fixupmessages: ...;is:=error`

PR #123, opened 2020-06-01
[codecov] Added code coverage

kriegfrj:

### DIFF
--- a/org.osgi.test.assertj/pom.xml
+++ b/org.osgi.test.assertj/pom.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-	* Copyright (c) OSGi Alliance (2019). All Rights Reserved.
+	* Copyright (c) OSGi Alliance (2019,2020). All Rights Reserved.
 	*
 	* Licensed under the Apache License, Version 2.0 (the "License");
 	* you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


org.osgi.test.assertj/pom.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-	* Copyright (c) OSGi Alliance (2019). All Rights Reserved.
+	* Copyright (c) OSGi Alliance (2019,2020). All Rights Reserved.
```
bjhargrave, line 3: There should be a space `2019, 2020` in these comments.

org.osgi.test.junit4/pom.xml
```diff
@@ -135,6 +135,10 @@
 					</execution>
 				</executions>
 			</plugin>
+			<plugin>
```
bjhargrave, line 138: This is already enabled in the parent pom.

kriegfrj:
> LGTM
> 
> I'm ok with this as long as it doesn't become more important to reach full coverage than to provide useful features.

Well, most of the other projects I've worked on (AssertJ, Mockito, Junit 5) have coffee coverage metrics,  i didn't want us to be left out.. :slightly_smiling_face: i also thought this is something much easier to have from the beginning than try and add later. 

I think of it more as a guide so that at a high level we can see where the gaps are and make sure we haven't left any accidentally. I've also found it useful for highlighting pockets of dead code.

So that it is not too onerous on us, i think the goal should be to not let coverage decrease rather than ensure we have 100%.

Issue #119, opened 2020-05-27
[commons] CloseableServiceObjects needs to keep track of use count


kriegfrj:
Was implementing #116 and wondering why my test kept failing - resources not getting cleaned up when expected. I think it's because I uncovered a real bug.

`CloseableServiceObjects` keeps track of services returned by `getService()` in an `IdentityHashMap`-backed set called `instances`. Then on close it iterates through them all and calls `ungetService()` to release them.

However, I discovered using the debugger that `getService()` internally increments the use count every time you call it, even if it returns the exact same service instance as a previous call. So really, you need to make sure that every `getService()` is paired with an `ungetService()`, and not simply call it once per service object.

Should be an easy fix (on its way).

Issue #118, opened 2020-05-22
[assertj] Soft assertions proxy generating references to non-imported packages


kriegfrj:
I started trying to use the newly-added soft proxies implementation of the `Bundle`, `BundleEvent` and `BundleContext` assertions. However, I found that the dynamically-generated soft assertions proxy classes have references to AssertJ core classes that are not statically referenced in the class files themselves. This means that bnd doesn't include them in `Import-Package` and the soft assertion class fails to load at runtime with `NoClassDefFoundError`. In my case, it was `org.assertj.core.api.recursive.comparison.RecursiveComparisonConfiguration`.

Possible solutions that I see:
* Manually add Import-Package filter to cover all of the core assertj packages (possibly as optional imports).
* Ditto but use DynamicImport-Package
* A Bnd plugin that can analyze the class files to find all instances of proxy generation invocation, run them, and examine the dependencies of the generated classes, and include those in the import-package analysis (Rolls-Royce!)

I tried the first two but only succeeded in getting rid of the default Import-Package statement. I tried using the syntax `Import-Package: org.assertj.core.api.recursive.comparison, *` only succeeded in ending up with an Import-Package statement that included only `org.assertj.core.api.recursive.comparison`. Ditto - when I tried `DynamicImport-Package: *`, the `Import-Package` statement disappeared completely.

kriegfrj:
Managed to figure out what i was doing wrong and now I've got the Import-Package working. Maybe we should also be doing full in-framework integration testing of our AssertJ bundle to catch these.

bjhargrave:
> Maybe we should also be doing full in-framework integration testing of our AssertJ bundle to catch these.

That is probably a necessary thing since the purpose of these is to use for in-framework integration testing. We should be confident that it works well.

Issue #116, opened 2020-05-20
[junit] Use junit-platform-testkit for testing extensions


kriegfrj:
I have learned in the last couple of years that mocking as a testing technique is fraught, especially as your level of functionality increases:

1. you spend as much time debugging your mocks as you spend debugging your actual code;
2. in spite of 1, your mocks are an approximation, and you can end up with tests that pass but the deliverable will still fail in the real world.

For this reason, I think it is almost always better to test against real code (where practical) rather than mocks.

Having wrestled briefly with #113 and #114, I think we have reached that point now where our mock-based tests are getting difficult to maintain and error-prone. Though it will involve a bit of work, I think we would be well served for the long time by putting in that effort now to port them to use `junit-platform-testkit`. I think that AssertJ's tests for `SoftAssertionsExtension` are a good model to follow (they were written by one of the core Jupiter developers): https://github.com/joel-costigliola/assertj-core/tree/master/src/test/java/org/assertj/core/api/junit/jupiter

kriegfrj:
And I forgot point 3, which was the one that actually prompted me to raise this issue: when low-level unit testing, you often end up testing behaviour that doesn't need to be tested. Not only is this a waste of time, it also means that when you refactor that code you break a bunch of all the lower-level tests even though the higher-level tests continue to pass, and you spend additional time fixing the broken test code.

My specific example: in `BundleContextExtension` we are manually cleaning up the state objects in the `afterEach()` callback. I changed the Namespace key in the beforeEach() implementation which meant that it was storing the `CloseableResourceBundleContext` in a different place to where `afterEach()` was looking for it, so `afterEach()` was effectively a no-op. Yet all the tests in `BundleContextExtensionExampleTest` were still passing (even those that I had added to verify that the context was being properly closed)! This is because `CloseableResourceBundleContext` implements the Jupiter API interface `CloseableResource`, and the contract for `CloseableResource` stipulates that Jupiter will automatically close any resources in the Store that implement `CloseableResource` when the Store is closed (ie, the ExtensionContext in which they were created goes out of scope and is closed). Realising that the code in `afterEach()` was redundant, I removed it - `BundleContextExtensionExampleTest` continued to pass, but many of the tests in `BundleContextExtensionTest`, `ServiceExtensionTest` etc started to fail because they were specifically testing for `afterEach()` to clean up.

bjhargrave:
I personally never really got into mock testing. Seems like too much voodoo magic to me :-)

[...]

Issue #114, opened 2020-05-18
[junit5] Extensions to properly handle parallelism


kriegfrj:
While having a look at #113, I realised that our current implementation of the JUnit 5 extensions and use of the Store is not parallel-safe. Because neither the key nor the namespace that we are using for the context lookup makes reference to the test method, every test method within the test class will always get the same BundleContext instance (or whatever state object it is storing). If their execution happens to overlap, this will cause them to interfere with each other.

The fix should be simple - to include the method name in the Namespace. See the `TimingExtension.getStore()` example in the JUnit 5 user doc [here](https://junit.org/junit5/docs/current/user-guide/#extensions-lifecycle-callbacks-before-after-execution)

I will look at this in conjunction with #113.

kriegfrj:
When I implemented my first set of test cases to test for this condition, I was surprised to find that they passed. On reflection, this makes sense as every method already gets its own unique ExtensionContext.

However, it might become an issue in conjunction with #113, because of the hierarchical lookup - if the Store doesn't find its key in the current context it will look in the parent context, and if it doesn't find one there _only then_ it will create a new one. Our implementation currently works because we never create our state object at the level of the test class, so the parent context never has one. But if I implement #113 then this might no longer apply.

It also occurred to me that in the current implementation, the bundle context state may persist between dynamic tests, because each dynamic test is the child of a test method, and so every dynamic invocation's context lookup will retrieve the parent test method's context. I haven't tested this theory yet but will shortly.

PR #126, opened 2020-06-03
[junit5] Multilevel extensions

kriegfrj:
This is a bit of a grab-bag of fixes:
* An implementation of #113 and #114 for `BundleContextExtension` (still a few fixes required for `ServiceExtension`.
* Implementation of #116 for JUnit 5 (could still do the same for JUnit 4 tests).
* Fix for #119 and #118.

The main feature of note is the hierarchical `BundleContextExtension`. It will inject bundle contexts at multiple levels in the scope hierarchy. Changes made are cleaned up once that level of the hierarchy goes out of scope. So for example:

* Declare a `BundleContext` at the class/`@BeforeAll` level or with a static scope and you can register services that will hang around and be visible to all tests in the test class, then get cleaned up once the last test is finished.
* Declare a `BundleContext` at the instance level or parameter injection can be used to register services for the life of the test method.
* Declare a `BundleContext` in an `@Nested`-annotated inner class and it will be visible to all methods within that class.

Comments welcome.


org.osgi.test.assertj/pom.xml
```diff
@@ -80,6 +80,20 @@
...
+				<executions>
+					<!-- Integration Test Configuration -->
+					<execution>
+						<id>bnd-process</id>
+						<goals>
+							<goal>bnd-process</goal>
+						</goals>
+						<configuration>
+							<bnd><![CDATA[
+								Import-Package: org.assertj.core.api.recursive.comparison,*
```
bjhargrave, line 92: Perhaps we should do this in source code instead of here? For example, I put the following in `AbstractBundleAssert` to cause the desired package import.

```java
	static { // SoftAssertions will require this at runtime
		Class<?> importRecursive = RecursiveComparisonConfiguration.class;
	}
```

See also https://github.com/osgi/osgi/issues/100 about making a better way to handle this in the future.

org.osgi.test.common/src/main/java/org/osgi/test/common/context/CloseableBundleContext.java
```diff
@@ -237,7 +237,11 @@ public ClosableServiceObjects(ServiceObjects<S> so) {
 
 		@Override
 		public void close() {
-			instances.forEach(so::ungetService);
+			instances.forEach((service, useCount) -> {
+				for (int i = 0; i < useCount; i++) {
```
bjhargrave, line 254: We should explicitly unbox `useCount`:

```java
for (int i = useCount; i > 0; i--) {
  so.ungetService(service);
}
```

org.osgi.test.common/src/main/java/org/osgi/test/common/context/CloseableBundleContext.java
```diff
@@ -258,19 +262,35 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
...
 		public S getService() {
 			S service = so.getService();
-			instances.add(service);
+			instances.merge(service, 1, (oldValue, dummy) -> oldValue + 1);
 			return service;
 		}
 
 		@SuppressWarnings("unused")
 		public void ungetService(S service) {
+			// FIXME: need to decrement instance count rather than remove completely
```
bjhargrave, line 306: Shouldn't this PR include this?

org.osgi.test.junit5/test.bndrun
```diff
@@ -53,4 +53,8 @@
 	org.osgi.test.common;version='[0.9.0,0.9.1)',\
 	org.osgi.test.junit5-tests;version='[0.9.0,0.9.1)',\
 	org.osgi.test.junit5;version='[0.9.0,0.9.1)',\
-	osgi.promise;version='[6.0.0,6.0.1)'
+	osgi.promise;version='[6.0.0,6.0.1)',\
+	junit-jupiter-params;version='[5.6.2,5.6.3)',\
```
bjhargrave, line 57: We have been trying to keep this in sorted order, so we should move the junit things up.

org.osgi.test.junit5/test.bndrun
```diff
@@ -53,4 +53,8 @@
...
+	junit-platform-testkit;version='[1.6.2,1.6.3)'
+
+-runee: JavaSE-1.8
```
bjhargrave, line 60: Do we need this?

org.osgi.test.assertj/pom.xml
kriegfrj: Sounds like a plan! Yes, that will make the dependency more obvious.

My gut feeling here actually is that the ultimate solution is to fix it in AssertJ somehow. Perhaps wrap it in a custom classloader that can delegate both to the bundle classloader for the bundle that hosts the original assertion, *and* also delegate to the bundle classloader for assertj-core. That will allow the proxy process to do whatever delegations/transformations it wishes to internally with the classes that are visible to it, without users of the bundle needing to figure out which classes they need to import by trial-and-error.

kriegfrj: -btw, that link to osgi/osgi#100 goes nowhere? Private repo maybe?

org.osgi.test.common/src/main/java/org/osgi/test/common/context/CloseableBundleContext.java
kriegfrj: Ok. (The compiler is not smart enough to do that?)

org.osgi.test.junit5/test.bndrun
kriegfrj: No - it snuck in when I was trying to debug bndtools/bnd#4130

org.osgi.test.assertj/pom.xml
bjhargrave: > Private repo maybe?

Duh. Yes. It will hopefully go public sooner rather than later. Sorry for the confusion.

The link suggests defining an Import annotation in OSGi that can be used to cause a package import. Basically it will make a "hard" reference to a package without actually having to write code to do it.

org.osgi.test.common/src/main/java/org/osgi/test/common/context/CloseableBundleContext.java
bjhargrave: Never trust a compiler when you can do the right thing easily yourself :-)

org.osgi.test.common/src/main/java/org/osgi/test/common/context/CloseableBundleContext.java
kriegfrj: I was planning to do a separate one. There is no code in the BundleContextExtension tests that calls ungetService() nor is there any need for one. I think the proper place for the test is in the commons bundle, which to my mind is a different PR (this one already had too much in it).

But I can look at this.

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/context/MultiLevelCleanupTest.java
```diff
@@ -0,0 +1,284 @@
...
+		setupResource();
+		assertSetup(BEFORE_CLASS, BEFORE_ALL, BEFORE_EACH, INNER_TEST);
+		assertNotSetup(PARAMETERIZED_TEST, NESTED_BEFORE_EACH, NESTED_TEST, NESTED_AFTER_EACH, AFTER_EACH, AFTER_ALL);
+	}
+
+	@ParameterizedTest
+	@ValueSource(ints = {
+		1, 2, 3
+	})
+	void parameterizedTest(int parameter) {
```
rotty3000, line 168: where is this parameter used?

org.osgi.test.junit5/bnd/afterAll.bnd
```diff
@@ -0,0 +1 @@
+Export-Package: ${p}.tb1.*;-split-package:=first
```
rotty3000, line 1: seems odd to have all these embedded jars have the exact same contents, no? Why no reuse just one?

org.osgi.test.junit5/pom.xml
```diff
@@ -136,9 +140,21 @@
 								-make: (*).(jar); type=bnd; recipe="${.}/bnd/$1.bnd"
 								-includeresource:\
 									tb1.jar,\
-									foo/tbfoo.jar=tb1.jar
+									beforeClass.jar,\
```
rotty3000, line 143: same question about all the embedded jars with the same contents. Do we really need them?

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/context/MultiLevelCleanupTest.java
rotty3000: ...or is this some trick to make the test execute multiple times? If so shouldn't `@RepeatedTest(3)` be used?

kriegfrj: In the test display name, and probably also internally in the test descriptor unique ID. It's not actually used in the test anywhere, but you can't have a parameterized test without parameters and I wanted to make sure the tests covered that case.

org.osgi.test.junit5/bnd/afterAll.bnd
kriegfrj: You can't install multiple bundles with the same bsn at the same time. In the multi-level test there are several different bundles installed at any given time.

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/context/BundleContextExtension_BundleContextInjectionTest.java
```diff
@@ -0,0 +1,142 @@
...
+			.isSameAs(bc)
+			.refersToBundle(bundle);
+		softly.assertAll();
+	}
+
+	@ParameterizedTest
+	@ValueSource(ints = {
+		1, 2, 3
+	})
+	void parameterizedTest(int parameter, @InjectBundleContext BundleContext bc) {
```
rotty3000, line 71: more unused parameters. what's the goal here?

kriegfrj: As noted above, the goal is to test that the extension works with different types of test nodes. Parameterized tests are one of the different types of test node, and you can't have a parameterized test without parameters.

org.osgi.test.junit5/pom.xml
rotty3000: comments?

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/context/BundleContextExtension_BundleContextInjectionTest.java
rotty3000: comments?

org.osgi.test.junit5/bnd/afterAll.bnd
kriegfrj: Where would you like them? In the bnd files or in the pom? Or both?

kriegfrj:
Latest push includes @bjhargrave 's review suggestions but not @rotty3000 's (yet).

org.osgi.test.assertj/pom.xml
bjhargrave: See https://github.com/bndtools/bnd/pull/4132

kriegfrj:
> There is no documentation about this change.

This is a good point. I normally like to leave this until after the review in case the implementation changes in a significant way which would require re-writing the documentation. But I will include it now.

org.osgi.test.junit5/bnd/afterAll.bnd
kriegfrj: The comment about it being opaque is fair. I will try and add some comments to explain the whole thing.

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/context/MultiLevelCleanupTest.java
kriegfrj: I have documented the unused parameter.

org.osgi.test.junit5/bnd/afterAll.bnd
kriegfrj: I added comments in the .bnd files that refer to the appropriate place in the code where they are used, and then explained in the code how/why they are used.

org.osgi.test.junit5/pom.xml
kriegfrj: Done, per above.

kriegfrj:
@rotty3000 , let me know if you're happy with the doc changes/if they make sense. Feel free to be critical - if they don't make sense to you then the average developer won't have a chance...

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/context/BundleContextExtensionExampleTest.java
```diff
@@ -46,7 +68,38 @@ public void testBundleContext1(@InjectBundleContext BundleContext bundleContext1
...
+			.isSameAs(
+				currentMethodContext)
+			.isNotSameAs(classLevelContext);
+	}
+
+	@ParameterizedTest
+	@ValueSource(ints = {
+		1, 2, 3
+	})
+	void testBundleContextFromDynamic(int value, @InjectBundleContext BundleContext bc) {
```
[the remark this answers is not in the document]
kriegfrj: Added some comments.

kriegfrj:
Given that the only complaints that @rotty3000 were around documentation rather than function, I have gone ahead and merged this request as I will need the functionality shortly. If there are any concerns or further identified deficiencies in the documentation, we can easily address those in future PRs.

Thanks for the reviews, gentlemen!

PR #161, opened 2020-07-03
[junit5] Sanity checking fixes

kriegfrj:
Factored out some sanity checking functions into `common`.
Added sanity checking tests for `ServiceExtension`.
Removed useless `public String NAMESPACE` from `BundleContextExtension`.

### DIFF
--- a/org.osgi.test.common/src/main/java/org/osgi/test/common/inject/FieldInjector.java
+++ b/org.osgi.test.common/src/main/java/org/osgi/test/common/inject/FieldInjector.java
@@ -35,6 +35,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.function.Predicate;
 
 public class FieldInjector {
@@ -57,6 +58,10 @@ public class FieldInjector {
 
 	public static List<Field> findAnnotatedNonStaticFields(Class<?> clazz, Class<? extends Annotation> annotationType) {
 		return findAnnotatedFields(clazz, annotationType, m -> !Modifier.isStatic(m.getModifiers()));
+	}
+
+	public static List<Field> findAnnotatedFields(Class<?> clazz, Class<? extends Annotation> annotationType) {
+		return findAnnotatedFields(clazz, annotationType, x -> true);
 	}
 
 	public static List<Field> findAnnotatedFields(Class<?> clazz, Class<? extends Annotation> annotationType,
@@ -206,6 +211,28 @@ public class FieldInjector {
 			.sorted(
 				fieldComparator)
 			.collect(toCollection(ArrayList::new));
+	}
+
+	public static void assertParameterIsOfType(Class<?> actual, Class<?> expected,
+		Class<? extends Annotation> annotationType, Function<String, ? extends RuntimeException> exception) {
+		if (actual != expected) {
+			throw exception.apply("Can only resolve @" + annotationType.getSimpleName()
+				+ " parameter of type " + expected.getName() + " but was: " + actual.getName());
+		}
+	}
+
+	public static void assertFieldIsOfType(Field field, Class<?> expected, Class<? extends Annotation> annotationType,
+		Function<String, ? extends RuntimeException> exception) {
+		if (field.getType() != expected) {
+			throw exception.apply(
+				"[" + field.getName() + "] Can only resolve @" + annotationType.getSimpleName() + " field of type "
+					+ expected.getName() + " but was: " + field.getType()
+						.getName());
+		}
+		if (Modifier.isFinal(field.getModifiers()) || Modifier.isPrivate(field.getModifiers())) {
+			throw exception.apply(
+				"@" + annotationType.getSimpleName() + " field [" + field.getName() + "] must not be private or final");
+		}
 	}
 
 }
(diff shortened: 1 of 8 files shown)


org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/context/package-info.java
```diff
@@ -15,5 +15,5 @@
  */
 
 @org.osgi.annotation.bundle.Export
-@org.osgi.annotation.versioning.Version("1.0.0")
+@org.osgi.annotation.versioning.Version("2.0.0")
```
bjhargrave, line 18: We don't need to bump the package version since we are still making `0` major version bundles. If the baselining against 0.9.0 bundles is complaining, we should disable baselining in the poms until we actually have a version 1.0.0 release.

org.osgi.test.common/src/main/java/org/osgi/test/common/inject/package-info.java
```diff
@@ -15,5 +15,5 @@
  */
 
 @org.osgi.annotation.bundle.Export
-@org.osgi.annotation.versioning.Version("1.0.0")
+@org.osgi.annotation.versioning.Version("1.1.0")
```
bjhargrave, line 18: We don't need to bump the package version since we are still making 0 major version bundles. If the baselining against 0.9.0 bundles is complaining, we should disable baselining in the poms until we actually have a version 1.0.0 release.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceExtension.java
```diff
@@ -114,31 +113,27 @@ public boolean supportsParameter(ParameterContext parameterContext, ExtensionCon
...
+		throw new ParameterResolutionException("Can only resolve @" + InjectService.class.getSimpleName()
+			+ " parameter for services with non-generic types, service type was: " + serviceType.getTypeName());
 	}
 
 	static void assertValidFieldCandidate(Field field) {
 		if (Modifier.isFinal(field.getModifiers()) || Modifier.isPrivate(field.getModifiers())
 			|| Modifier.isStatic(field.getModifiers())) {
-			throw new ExtensionConfigurationException(
-				InjectService.class.getName() + " field [" + field + "] must not be final, private or static.");
+			throw new ExtensionConfigurationException('@' + InjectService.class.getSimpleName() + " field ["
```
bjhargrave, line 130: `"@"` since we are concatenating strings, we should use String constants. Not a char constant which must be converted to a String object at runtime.

org.osgi.test.common/src/main/java/org/osgi/test/common/inject/FieldInjector.java
```diff
@@ -208,4 +213,26 @@ static boolean isFieldShadowedByLocalFields(Field field, List<Field> localFields
...
+		Function<String, ? extends RuntimeException> exception) {
+		if (field.getType() != expected) {
+			throw exception.apply(
+				"[" + field.getName() + "] Can only resolve @" + annotationType.getSimpleName() + " field of type "
+					+ expected.getName() + " but was: " + field.getType()
+						.getName());
+		}
+		if (Modifier.isFinal(field.getModifiers()) || Modifier.isPrivate(field.getModifiers())) {
+			throw exception.apply(
+				'@' + annotationType.getSimpleName() + " field [" + field.getName() + "] must not be private or final");
```
bjhargrave, line 234: `"@"` since we are concatenating strings, we should use String constants. Not a char constant which must be converted to a String object at runtime.

org.osgi.test.junit5/test.bndrun
```diff
@@ -24,6 +24,9 @@
 	org.osgi.service.http.port=*,\
 	org.osgi.framework.bootdelegation=sun.reflect,\
 	osgi.console=
+#	,\
+#	tester.names=org.osgi.test.junit5.service.ServiceExtension_SanityCheckingTest
```
bjhargrave, line 28: Do we really want to commit this? This seems like something you used to test your changes?

kriegfrj: Err, no. I was clearly in too much of a hurry to commit. :smile: Sorry. 

The idea of that line was so it would only run the test i was working on, speeds up development.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/context/package-info.java
kriegfrj: Fixed.

org.osgi.test.common/src/main/java/org/osgi/test/common/inject/package-info.java
kriegfrj: Fixed.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceExtension.java
kriegfrj: Fixed.

I didn't believe it until I looked at the decompiled output - I had assumed that the compiler would generate a call `StringBuilder.append(char)` rather than `StringBuilder.append(String.valueOf(char))`. Why doesn't it???

org.osgi.test.common/src/main/java/org/osgi/test/common/inject/FieldInjector.java
kriegfrj: Fixed.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceExtension.java
bjhargrave: > Why doesn't it???

🤷‍♂️

Issue #164, opened 2020-07-17
Provide way to handle configAdmin configutations via Annotations/Extension


stbischof:
The ConfigurationExtension should help to config the Config-Admin using annotations. I also should inject Fields and Parameters of the `org.osgi.service.cm.Configuration` Type in annotated.

Could you have a look at this Example and give me some feedback
```
@Testable
@ExtendWith(ServiceExtension.class)
@ExtendWith(ConfigurationExtension.class)
@Configuration(pid = "my.pidClass1", factory = true, properties = {
        @ConfigEntry(key = "foo", value = "bar") })
@Configuration(pid = "my.pidClass2", factory = true, properties = {
        @ConfigEntry(key = "key1", value = "foo"),
        @ConfigEntry(key = "key2", value = "bar") })
public class ConfigAnnotationTest
{
    //Field
    @Configuration(pid = "my.pidField", factory = true, properties = {
            @ConfigEntry(key = "key1", value = "foo"),
            @ConfigEntry(key = "key2", value = "bar") })
    org.osgi.service.cm.Configuration configuration;

    //Method
    @Test
    @Configuration(pid = "my.pid2", factory = true, properties = {
            @ConfigEntry(key = "key1", value = "value"),
            @ConfigEntry(key = "key2", value = "value") })
    public void singleConfigOnMethod()
    {
        System.out.println();
    }

    @Test
    @Configuration(pid = "my.pid1", factory = true, properties = {
            @ConfigEntry(key = "key", value = "value") })
    @Configuration(pid = "my.pid2", factory = true, properties = {
            @ConfigEntry(key = "key", value = "value") })
    public void multiConfigOnMethod()
    {
        System.out.println();
    }

    @Test
    @Configurations(value = {
            @Configuration(pid = "my.pid1", factory = true, properties = {
                    @ConfigEntry(key = "key", value = "value") }),
            @Configuration(pid = "my.pid2", factory = true, properties = {
                    @ConfigEntry(key = "key", value = "value") }) })
    public void multiConfigOnMethodWrapped()
    {
        System.out.println();
    }

    @Test
    public void intectConfigOnParameter(
        @Configuration(pid = "my.pid1", factory = true, properties = {
                @ConfigEntry(key = "key1", value = "foo"),
                @ConfigEntry(key = "key2", value = "bar") }) org.osgi.service.cm.Configuration configuration)
    {
        System.out.println();
    }

    private final String PID_NAME = "my.pid";

    @Test
[...]

PR #165, opened 2020-07-18
Provide a way to have configuration in ConfigurationAdmin before Test

stbischof:
ref #164


org.osgi.test.junit5/pom.xml
```diff
@@ -99,6 +99,10 @@
 			<groupId>org.osgi</groupId>
 			<artifactId>org.osgi.service.log</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.osgi</groupId>
+			<artifactId>org.osgi.service.cm</artifactId>
```
bjhargrave, line 104: I don't think we want to make the test artifacts from this projects dependent on many different OSGi API packages. This PR makes org.osgi.test.junit5  dependent on Config Admin at runtime (it will need to import the package). This makes org.osgi.test.junit5 awkward to use when I am not even testing or using CM. I think we will need to make individual artifacts for each new OSGi API that we want to test. Similar to how we have different artifacts for the assertj tests.

pom.xml
```diff
@@ -401,6 +401,12 @@
 				<version>6.0.0</version>
 				<scope>provided</scope>
 			</dependency>
+			<dependency>
+				<groupId>org.osgi</groupId>
+				<artifactId>org.osgi.service.cm</artifactId>
+				<version>1.6.0</version>
+				<scope>provided</scope>
```
bjhargrave, line 408: This is clearly NOT provided. The OSGI framework does not export the cm package at runtime.

stbischof:
I thought i may could be enough to set the requirement as optional

```
@Requirement(namespace = "osgi.implementation", //
	name = ConfigurationConstants.CONFIGURATION_ADMIN_IMPLEMENTATION, //
	version = ConfigurationConstants.CONFIGURATION_ADMIN_SPECIFICATION_VERSION, resolution = Resolution.OPTIONAL
```


What should be the name of the new bundle?

org.osgi.test.junit5.core
org.osgi.test.junit5.cm
...?

bjhargrave:
> I thought i may could be enough to set the requirement as optional

Well that says nothing about the package requirement (Import-Package). And the test code needs to have a mandatory requirement on the minimum version of Config Admin which the test code supports.

> What should be the name of the new bundle?

I would suggest to leave `org.osgi.test.junit5` (already shipped in 0.9.0) the same and create a new `org.osgi.test.junit5.cm` artifact.

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/Configurations.java
```diff
@@ -0,0 +1,23 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Inherited
+@Target({
+	TYPE, METHOD, PARAMETER, FIELD
+})
+@Retention(RUNTIME)
+@Documented
+public @interface Configurations {
```
kriegfrj, line 20: I've not had much experience practically with defining repeatable annotations, but I just found this little tip which I thought might be useful and help tidy up the code a little - basically, to make the container annotation an inner class:

https://dzone.com/articles/repeatable-annotations-in-java-8-1

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/Configuration.java
```diff
@@ -0,0 +1,46 @@
...
+import java.lang.annotation.Target;
+
+@Inherited
+@Target({
+	TYPE, METHOD, PARAMETER, FIELD
+})
+@Repeatable(Configurations.class)
+@Retention(RUNTIME)
+@Documented
+public @interface Configuration {
```
kriegfrj, line 38: If we follow the pattern used in our other injector extensions, we should call this `@InjectConfiguration` rather than simply `@Configuration`. This would also help to avoid the name collision with `org.osgi.service.cm.Configuration`, which I can forsee getting annoying for users of this extension.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/config/ConfigurationExtension.java
```diff
@@ -0,0 +1,156 @@
...
+						});
+				}
+				Configuration configAnnotation = element.getAnnotation(Configuration.class);
+				if (configAnnotation != null) {
+					resolveReturnValue(configAnnotation, bundleContext, extensionContext);
+				}
+			});
+	}
+
+	private static org.osgi.service.cm.Configuration resolveReturnValue(Configuration configurationAnnotation,
```
kriegfrj, line 86: If you follow my other suggestion and use `@InjectConfiguration` instead of `@Configuration` for the annotation, then that will eliminate the name collision and you won't need to use the FQN, which would make this a bit cleaner. I think generally it will be common for users of this code to want to refer to both so it will make all the users of this code cleaner too.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/config/ConfigurationExtension.java
```diff
@@ -0,0 +1,156 @@
...
+						.forEachOrdered((configAnnotation) -> {
+							resolveReturnValue(configAnnotation, bundleContext, extensionContext);
+						});
+				}
+				Configuration configAnnotation = element.getAnnotation(Configuration.class);
+				if (configAnnotation != null) {
+					resolveReturnValue(configAnnotation, bundleContext, extensionContext);
+				}
+			});
+	}
```
kriegfrj, line 84: I've recently tidied up and factored out some of the equivalent code in `BundleContextExtension` - have a look and see if you can reuse it. Some of the code now lives here: https://github.com/osgi/osgi-test/blob/6a756341e12349faebd61a43acb14a38e7368cc8/org.osgi.test.common/src/main/java/org/osgi/test/common/inject/FieldInjector.java#L216-L236

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/config/ConfigAnnotationTest.java
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	@Test
+	public void testParameterConfiguration2(@Configuration(pid = PARAM_PID, properties = {
+		@ConfigEntry(key = "foo", value = "bar")
+	}) org.osgi.service.cm.Configuration configuration) throws Exception {
+
+		List<org.osgi.service.cm.Configuration> cs = getConfigsByServicePid(PARAM_PID);
+		assertEquals(1, cs.size());
+		assertEquals(cs.get(0), configuration);
```
kriegfrj, line 118: AssertJ is much better at this than standard JUnit 5 assertions: `assertThat(cs).containsExactly(configuration)` Not only is the code more compact and readable, the diagnostic message (if the assertion fails) is much better too.

I highly recommend using AssertJ for these tests - we've used it for most of the unit tests in this project. Additionally, if you're going to be doing multiple assertions per test, you should look at using soft assertions (perhaps with the `SoftAssertionsExtension` to make your life easier).

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/config/ConfigurationExtension.java
```diff
@@ -0,0 +1,156 @@
...
+
+	@Override
+	public boolean supportsParameter(ParameterContext parameterContext, ExtensionContext extensionContext)
+		throws ParameterResolutionException {
+
+		if (parameterContext.isAnnotated(Configuration.class)) {
+			return true;
+		}
+		return false;
+	}
```
kriegfrj, line 154: Best practice (I'm using AssertJ's `SoftAssertionsExtension` for reference here as it was written by one of the JUnit 5 team) is for `supportsParameter` to also do the sanity checking if you know it's supposed to be one of your parameters, and throw a `ParameterResolutionException` if it doesn't pass validation. Again, have a look at `BundleContextExtension` for an example of how to do this (and how to write unit tests for it too).

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/ConfigEntry.java
```diff
@@ -0,0 +1,39 @@
...
+})
+@Retention(RUNTIME)
+@Documented
+public @interface ConfigEntry {
+
+	String key();
+
+	String value() default "";
+
+}
```
kriegfrj, line 39: I'm not sure if this annotation adds much value. Given that it only accepts strings anyway, it would probably be easier if the `@Configuration.properties()` was simply a `String[]`, with each string of the form "name=value". This is how properties are specified using DS injection, for example.

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/Configurations.java
bjhargrave: I don't find the idea of an inner type that compelling. We did not do this for repeatable OSGi annotations.

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/ConfigEntry.java
bjhargrave: Using `"name=value"` is kind of a hack. In DS we now prefer ComponentPropertyType since the annotation is user defined and type safe. But that may be a lot for testing to support that. The one thing missing in ConfigEntry is to support the property type since CM supports more that String values.

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/Configurations.java
kriegfrj: I defer to the (more experienced) majority.

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/ConfigEntry.java
kriegfrj: Using annotations for type safety would make sense, but as you noted at the moment `@ConfigEntry` only supports strings. If we're going to use the annotation so we can get type safety in the future, a bit of a simple roadmap might be a good idea. 

This is actually a similar problem space to the sources that Jupiter's parameterised tests use. Perhaps we could look there for inspiration.

stbischof:
Im not able to tun the test within the itest. Could you help me?

org.osgi.test.junit5.cm/pom.xml
```diff
@@ -0,0 +1,253 @@
...
+						<configuration>
+							<artifactFragment>true</artifactFragment>
+							<bnd><![CDATA[
+							Test-Cases: org.osgi.test.junit5.cm.ConfigAnnotationTest
+							]]></bnd>
+<!-- 							Test-Cases:  -->
+<!--                            Test-Cases: org.osgi.test.junit5.cm.ConfigAnnotationTest -->
+							<includeClassesDir>false</includeClassesDir>
+							<testCases>useTestCasesHeader</testCases>
+<!--                             <testCases>${classes;HIERARCHY_INDIRECTLY_ANNOTATED;org.junit.platform.commons.annotation.Testable;CONCRETE;PUBLIC}</testCases> -->
```
bjhargrave, line 160: This should be `<testCases>junit5</testCases>` (which is the default anyway). You cannot specify a value other then the enum values.

stbischof: thank you. eclipse build a bar with the class...

kriegfrj: Heh, exactly the same issue that caught me out in bndtools/bnd#4088 If we can't easily sanity check for this in the Mojo, then perhaps we should make a loud warning about it in the documentation.

kriegfrj: Scratch that - I remember that I deliberately *didn't* use the default junit5 enum. There was a reason for this - the default enum includes all concrete classes in `Test-Cases`, however the JUnit 5 test bundle has lots of test case classes in that that are used as "fodder" for `EngineTestKit`. We don't want maven/Bndtools to run these fodder test cases directly. My way around it was to use a custom Test-Cases header which only included the *public* test cases and ignored the package-private ones (the default macro used by junit5 includes public and package-private classes). This allowed me to declare the fodder classes as package-private so that they wouldn't get included in `Test-Cases`. But then you have to explicitly declare the `Test-Cases` directive as part of the `<bnd>` parameter.

stbischof:
Could you please have a look on the  [InjectConfiguration.java ](https://github.com/osgi/osgi-test/pull/165/files#diff-b925ece862582db2eb4570e4b5b46e45R31-R85).
Any ideas for a better methodNames or what else could be handles with this Annotation?

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/InjectConfiguration.java
```diff
@@ -0,0 +1,86 @@
...
+	String pid();
+
+	/**
+	 * Indicate the name of the Factory-Configuration (not {@code null}). If
+	 * empty - auto-generated UUID will be used.
+	 *
+	 * @return The name of the Factory.
+	 */
+
+	String factory_name() default "";
```
kriegfrj, line 61: I think we've used camelCase elsewhere? We should be consistent.

stbischof: Ok! Like filterArguments in InjectService.

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationNested.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.osgi.service.cm.Configuration;
+import org.osgi.test.common.annotation.config.CmAction;
+import org.osgi.test.common.annotation.config.InjectConfiguration;
+
+@ExtendWith(ConfigurationExtension.class)
+@InjectConfiguration(pid = "sub.0")
```
bjhargrave, line 13: Where is this configuration being injected?

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationNested.java
```diff
@@ -0,0 +1,37 @@
...
+@Nested
+public class ConfigAnnotationNested {
+
+	@InjectConfiguration(pid = "sub.1")
```
bjhargrave, line 17: Where is this configuration being injected?

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationNested.java
```diff
@@ -0,0 +1,37 @@
...
+	@Nested
+	class SubClass {
+
+		@InjectConfiguration(pid = "sub.2")
```
bjhargrave, line 21: Where is this configuration being injected?

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
```diff
@@ -127,7 +129,7 @@ public void testMethodConfiguration() throws Exception {
 	}
 
 	@Test
-	@InjectConfiguration(pid = FACTORY_CONFIGURATION_PID, factory_name = "factory.name", factory = true, properties = {
+	@InjectConfiguration(pid = FACTORY_CONFIGURATION_PID, factoryName = "factory.name", factory = true, properties = {
```
bjhargrave, line 132: Where is this configuration being injected?

stbischof:
That is why the annotation at the beginning was `@Configuration`. 

The `@Configuration` / `@InjectConfiguration` would have the same function.
- get or create a Configuration
- update the Configuration

Only if it is on an Field or Parameter it should additionally also inject it.

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationNested.java
stbischof: "into the framework;)". So you can separate test configurations using nested tests with inner classes

org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/package-info.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (c) OSGi Alliance (2019). All Rights Reserved.
```
kriegfrj, line 2: Minor nit: there are a bunch of cut-and-paste errors here - some are (2019), others are (2019, 2020). All of them in this PR should be just (2020).

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+		ca = bundleContext.getService(sref);
+		if (ca == null) {
+			throw new IllegalStateException("The configuration admin service cannot be found.");
+		}
+
+	}
+
+	@AfterAll
+	public static void afterEach() {
```
kriegfrj, line 48: Either you've used the wrong annotation here, or you've used a confusing name for your method. :smile: I'm guessing you meant this method to be called "afterAll()".

org.osgi.test.junit5.cm/test.bndrun
```diff
@@ -0,0 +1,55 @@
...
+	junit-platform-engine;version='[1.6.2,1.6.3)',\
+	junit-platform-launcher;version='[1.6.2,1.6.3)',\
+	org.opentest4j;version='[1.2.0,1.2.1)',\
+	org.osgi.test.common;version='[0.10.0,0.10.1)',\
+	org.osgi.test.junit5.cm-tests;version='[0.10.0,0.10.1)',\
+	org.osgi.service.cm;version='[1.6.0,1.6.1)',\
+	assertj-core;version='[3.16.1,3.16.2)',\
+	org.apache.felix.configadmin;version='[1.9.16,1.9.17)',\
+	org.osgi.test.junit5.cm;version='[0.10.0,0.10.1)'
+-runee: JavaSE-11
```
kriegfrj, line 55: We shouldn't set this - Maven sets it based on the Maven environment (or if you must set it for some reason, it should probably be set to JavaSE-1.8). This is what the other integration tests due.

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+	static final String PARAM_PID = "param.pid";
+
+	@Test
+	public void testParameterConfiguration(@InjectConfiguration(pid = PARAM_PID, properties = {
+		@ConfigEntry(key = "bar", value = "foo")
+	}) Configuration configuration) throws Exception {
+
+		Configuration cs = ConfigAdminUtil.getConfigsByServicePid(ca, PARAM_PID);
+		assertThat(cs).isEqualTo(configuration);
```
kriegfrj, line 78: Should we perhaps assert here `isSameAs()` rather than merely `isEqualTo()`? Or do you envisage a valid use case where the injected configuration might be a different instance to the one returned by the `ConfigAdminUtil.getConfigsByServicePid()`?

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
```diff
@@ -0,0 +1,146 @@
...
+		assertEquals("foo", foo);
+	}
+
+	@Test
+	public void testParameterConfiguration2(@InjectConfiguration(pid = PARAM_PID, properties = {
+		@ConfigEntry(key = "foo", value = "bar")
+	}) Configuration configuration) throws Exception {
+
+		Configuration cs = ConfigAdminUtil.getConfigsByServicePid(ca, PARAM_PID);
+		assertThat(cs).isEqualTo(configuration);
```
kriegfrj, line 94: `isSameAs()`?

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/InjectConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+	String factoryName() default "";
+
+
+	/**
+	 * Indicate that the Configuration is a Factory-Configuration .
+	 *
+	 * @return The factory-marker.
+	 */
+
+	boolean factory() default false;
```
kriegfrj, line 70: The combination of these two parameters makes it possible for the user to set an inconsistent configuration - if you specify a `factoryName` but set `factory=false` (eg, you forget to set factory), what happens? One of two things should happen - either we throw an exception to alert the user to the inconsistency, or else we simply define that if `factoryName` is set to a non-empty value then `factory` is ignored.

Alternatively, it might make more sense to have `factoryName` and `autoFactory`, and then throw an exception if `factoryName` is non-empty and `autoFactory` is `true`.

org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/ConfigurationExtension.java
```diff
@@ -0,0 +1,200 @@
...
+		return resolveReturnValue(configAnnotation.get(), bundleContext, extensionContext);
+	}
+
+	@Override
+	public boolean supportsParameter(ParameterContext parameterContext, ExtensionContext extensionContext)
+		throws ParameterResolutionException {
+
+		if (parameterContext.isAnnotated(InjectConfiguration.class)) {
+			return true;
+		}
```
kriegfrj, line 164: I'd still like to see some error checking here. If a field is annotated with `@InjectConfiguration` (or whatever the final name ends up being) but the type of the field is not `Configuration`, then we should throw an exception here.

kriegfrj:
> That is why the annotation at the beginning was `@Configuration`.
> 
> The `@Configuration` / `@InjectConfiguration` would have the same function.
> 
> * get or create a Configuration
> * update the Configuration
> 
> Only if it is on an Field or Parameter it should additionally also inject it.

This makes sense, as you won't always need to refer directly to the configuration in your test - you just need it to be present for your code-under-test. Seems like a waste of coding effort (and a compiler warning in the waiting) to inject it into a field that you're never going to explicitly access.

> Maybe we need a better name. I do not think that we should spit the annotation.
>

How about `@WithConfiguration`?

If we rename this and don't go for two annotations, we'll also need to think about whether or not we adjust the other extensions' annotations for consistency. Either we run two naming conventions in parallel, or we consider renaming `@InjectService`/`@InjectBundleContext`/`@InjectInstallBundle` to `@WithService` (etc).

stbischof:
When we think about naming, there is also a way to use ArgumentSources in ParameterizedTests.
This is what I testes yesterday #170  (Only the Bundle-Test is runs correctl).
At the moment I'm not sure that we really need this. But maybe we should also take care about the naming.

org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/ConfigurationExtension.java
stbischof: Ok,
If the `@InjectConfiguration` is set on Dictionaries or Maps, should we Inject  the properties of the configuration?

org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/package-info.java
stbischof: ok

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
stbischof: isSame is correct

stbischof: Surprise, we have to use Equals.

Felix ConfigurationAdminImpl creates a new ConfigurationAdapter each call of listConfigurations
```
    private Configuration wrap( ConfigurationImpl configuration )
    {
        return new ConfigurationAdapter( this, configuration );
    }
```

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/InjectConfiguration.java
stbischof: an other way would be to remove both annotations

`service.pid` - not a factory
`service.pid~` - a factory - auto generated
`service.pid~factoryName` - a factory with given factoryName

stbischof: @bjhargrave @kriegfrj could you have a look a this.

stbischof:
@kriegfrj @bjhargrave 
Could we find some naming-rules for annotations?

org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/ConfigurationExtension.java
kriegfrj: If that's not too hard to do, it might be useful. Both would be useful, but Map would probably be the more useful of the two because it wouldn't be too hard to inject a `Configuration` object and then simply call `Configuration.getProperties()`.

kriegfrj:
> @kriegfrj @bjhargrave
> Could we find some naming-rules for annotations?

Unless @bjhargrave or @rotty3000 strongly disagree, then I think for this PR let's just go with `@WithConfiguration`. Let's worry about if/when we change the other extensions to match in another PR after it's had a bit of use (we've got until our 1.0.0 to bed that down).

kriegfrj:
Sorry to trouble you, but can you also please add junit5.cm to the codecov upload in [`cibuild.yml`](https://github.com/osgi/osgi-test/blob/main/.github/workflows/cibuild.yml)? I'd like to see how we're going on our test coverage. We're not pedantic about getting to 100% coverage, but I would like to ensure that our test coverage is not decreasing overall.

org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/ConfigurationExtension.java
stbischof: done

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationNested.java
stbischof: Changes into @WithConfiguration

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/ConfigEntry.java
stbischof: added a type to the annotation

bjhargrave:
I like the `Inject*` annotations use for declaring the injection site and what should be injected. I think we need to now conflate creating a configuration with injecting it. 

Perhaps we need 2 annotations. One for creating a configuration (e.g. `WithConfiguration`) and another to inject a configuration (e.g. `InjectConfiguration`) .`InjectConfiguration` could take a String pid argument to get the named configuration from ConfigAdmin, or it could take a `WithConfiguration` argument which would mean create the configuration and _then_ inject it.

I can see the sometimes we need to create but not inject configurations. Sometimes we need to inject but not create configurations. And sometimes we need both. So two cooperating annotations make sense, no?

rotty3000:
I have to agree with @bjhargrave on having two annotations; `@InjectConfiguration` and `@WithConfiguration`. I like the separation but I also think we need to expose a timeout on `@InjectConfiguration` because it won't be a create-just-in-time type of scenario, which I think is perfectly valid (perhaps some code under test already creates configurations and you need to test they are correct.) With `@InjectService` there's an adjustable timeout with default. We should mirror that ability here.

stbischof:
@rotty3000 I am not sure what the best way to handle the timeout. Updates and deletes of configurations are async.

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/ConfigEntry.java
```diff
@@ -0,0 +1,41 @@
...
+})
+@Retention(RUNTIME)
+@Documented
+public @interface ConfigEntry {
+
+	String key();
+
+	String value() default "";
+
+	String type() default "String";
```
bjhargrave, line 39: I think this should should be an inner enum. See https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#i3217016 for the allowable types in a configuration.

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/InjectConfiguration.java
```diff
@@ -0,0 +1,63 @@
...
+	 * <br>
+	 * <ul>
+	 * <li><code>pid</code> - a service not a factory
+	 * <li><code>pid~factoryName</code> - a factory with given factoryName
+	 * </ul>
+	 *
+	 * @return The pid
+	 */
+	String value();
+
```
bjhargrave, line 56: Should we not also allow a `WithConfiguration` argument? So one can say

```java
@InjectConfiguration(@WithConfiguration(...))
```

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/WithConfigurationUtil.java
```diff
@@ -0,0 +1,56 @@
...
+	public static boolean isFactory(WithConfiguration withConfiguration) {
+		return isFactory(withConfiguration.pid());
+	}
+
+	public static boolean isFactory(InjectConfiguration injectConfiguration) {
+		return isFactory(injectConfiguration.value());
+	}
+
+	public static boolean isFactory(String pid) {
+		return pid.contains("~");
```
bjhargrave, line 54: Why does a factory configuration require `~`? This is just a convention for ConfigAdmin to generate a configuration pid from a factory pid and a name.

org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/ConfigurationCopy.java
```diff
@@ -0,0 +1,50 @@
...
+
+	private ConfigurationCopy() {
+
+	}
+
+	static ConfigurationCopy of(Configuration configuration) {
+
+		ConfigurationCopy copy = new ConfigurationCopy();
+		copy.pid = configuration.getPid();
+		copy.properties = configuration.getProperties();
```
bjhargrave, line 23: Shouldn't this be a deeper copy? Not just a reference?

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/WithConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+
+	/**
+	 * The pid of the Configuration.<br>
+	 * <br>
+	 * <ul>
+	 * <li><code>pid</code> - a service not a factory
+	 * <li><code>pid~</code> - a factory with auto-generated UUID will be used
+	 * as factoryName
+	 * <li><code>pid~factoryName</code> - a factory with given factoryName
+	 * </ul>
```
bjhargrave, line 55: This list seems a strange, new convention. Perhaps we need two annotations here: `WithConfiguration` for non-factory configurations and `WithFactoryConfiguration` for factory configurations. Then it is quite clear from the code what you are doing.

org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/ConfigurationCopy.java
stbischof: sure. will add a copy method into Dictionaries.class

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/WithConfigurationUtil.java
stbischof: removed most of the "~" magic

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/ConfigEntry.java
stbischof: See: https://github.com/osgi/osgi-test/pull/165/commits/4b57e835d598bdbdc72f515e916dbc74fd1f76be#diff-955bbd7d7884b63d979a2386d67dac99

https://github.com/osgi/osgi-test/pull/165/commits/4b57e835d598bdbdc72f515e916dbc74fd1f76be#diff-8abbc06a0596380453c03014d8e0ca12R454-R532

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/WithConfiguration.java
stbischof: very usefull:
https://github.com/osgi/osgi-test/pull/165/commits/4b57e835d598bdbdc72f515e916dbc74fd1f76be#diff-8af77c76c4aee96acb727d95a29fbca5

org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/ConfigurationCopy.java
stbischof: i copy the Dictionary. but maybe there is an better way to do this

https://github.com/osgi/osgi-test/pull/165/commits/4b57e835d598bdbdc72f515e916dbc74fd1f76be#diff-8abbc06a0596380453c03014d8e0ca12R448-R452

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/InjectConfiguration.java
stbischof: now you can use
@InjectConfiguration("pid")
@InjectConfiguration(withConfig=@WithConfiguration(...))
@InjectConfiguration(withFactoryConfig=@WithFactoryConfiguration(...))

org.osgi.test.junit5/test.bndrun
```diff
@@ -48,9 +48,6 @@
 	junit-platform-engine;version='[1.6.2,1.6.3)',\
```
bjhargrave, line 48: Remove this file from your PR. I am fixing it in https://github.com/osgi/osgi-test/pull/172.

org.osgi.test.junit5.cm/test.bndrun
```diff
@@ -0,0 +1,54 @@
...
+	ch.qos.logback.classic,\
+	ch.qos.logback.core,\
+	org.apache.felix.logback,\
+	slf4j.api
+-runrequires: \
+	bnd.identity;id='${project.artifactId}-tests',\
+	bnd.identity;id='junit-jupiter-engine',\
+	bnd.identity;id='junit-platform-launcher'
+# Please try and keep this sorted to make diffing more consistent
+-runbundles: \
```
bjhargrave, line 42: Please sort the runbundles list. We try to keep it in sorted order so that we can easily see what changes.

org.osgi.test.common/src/main/java/org/osgi/test/common/dictionary/Dictionaries.java
```diff
@@ -441,6 +445,92 @@ public String toString() {
 			.collect(toDictionary());
 	}
 
+	public static <K, V> Dictionary<K, V> copy(Dictionary<K, V> dictionary) {
```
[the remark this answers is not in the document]
stbischof: moved to junit5.cm.ConfigUtils

org.osgi.test.common/src/main/java/org/osgi/test/common/dictionary/Dictionaries.java
```diff
@@ -35,6 +36,9 @@
 import java.util.stream.Collector;
 import java.util.stream.Collectors;
 
+import org.osgi.test.common.annotation.config.ConfigEntry;
+import org.osgi.test.common.annotation.config.ConfigEntry.Type;
+import org.osgi.test.common.annotation.config.WithConfiguration;
```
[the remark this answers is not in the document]
stbischof: moved to junit5.cm.ConfigUtils

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/InjectConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/*
+ * Reads existing a Configuration from the Config-Admin
+ * Inject the `org.osgi.service.cm.Configuration` (if used on a Parameter or Field) for further updates or delete of the `org.osgi.service.cm.Configuration`
+ *
```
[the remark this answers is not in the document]
stbischof: done. in docs and throws IllegalArgumentException

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/WithConfigurationUtil.java
```diff
@@ -0,0 +1,7 @@
+package org.osgi.test.common.annotation.config;
+
+public class WithConfigurationUtil {
```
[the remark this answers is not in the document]
stbischof: removed

org.osgi.test.common/src/test/java/org/osgi/test/common/annotation/config/ConfigEntryTest.java
```diff
@@ -0,0 +1,11 @@
+package org.osgi.test.common.annotation.config;
```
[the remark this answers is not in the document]
stbischof: removed

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/ConfigEntry.java
```diff
@@ -0,0 +1,63 @@
...
+
+	String key();
+
+	String[] value() default "";
+
+	Scalar scalar() default Scalar.String;
+
+	Type type() default Type.Scalar;
+
+	boolean primitive() default false;
```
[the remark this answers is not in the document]
stbischof: thank you. have changed it

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/WithConfigurationUtil.java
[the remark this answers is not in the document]
bjhargrave: Still here :-)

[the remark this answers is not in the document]
stbischof: removed, really!

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationNested.java
```diff
@@ -0,0 +1,135 @@
...
+	@Test
+	@WithConfiguration(pid = "class.method1")
+	public void test_class_method1(@InjectConfiguration("class") Configuration c_class,
+		@InjectConfiguration("class.method1") Configuration c_class_Method1,
+		@InjectConfiguration("class.inner1") Configuration c_class_Inner1,
+		@InjectConfiguration("class.inner1.method1") Configuration c_class_Inner1_Method1,
+		@InjectConfiguration("class.inner1.method2") Configuration c_class_Inner1_Method2,
+		@InjectConfiguration("class.inner2") Configuration c_class_Inner2,
+		@InjectConfiguration("class.inner2.method1") Configuration c_class_Inner2_Method1) throws Exception {
+		assertNotNull(c_class);
```
bjhargrave, line 51: All the tests in this project would be much better if we used assertj assertions to test.

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+		bundleContext = FrameworkUtil.getBundle(ConfigAnnotationTest.class)
+			.getBundleContext();
+
+		sref = bundleContext.getServiceReference(ConfigurationAdmin.class);
+		if (sref == null) {
+			throw new IllegalStateException("The configuration admin service cannot be found.");
+		}
+
+		ca = bundleContext.getService(sref);
```
bjhargrave, line 58: We should be use ServiceExtention to get services for testing :-) Eat our own food.

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
```diff
@@ -0,0 +1,182 @@
...
+	@Test
+	@WithConfiguration(pid = PARAM_PID, properties = {
+		@ConfigEntry(key = "foo", value = "bar")
+	})
+	public void testParameterConfiguration2(
+		@InjectConfiguration(PARAM_PID) Configuration configuration)
+		throws Exception {
+
+		Configuration cs = ConfigUtil.getConfigsByServicePid(ca, PARAM_PID);
+		assertThat(cs).isEqualTo(configuration);
```
bjhargrave, line 118: Configuration.equals just compares the configuration pids. Not the configuration properties. Is that what you mean here?

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
```diff
@@ -0,0 +1,182 @@
...
+
+	@Test
+	@WithConfiguration(pid = PARAM_PID)
+	public void testParameterConfiguration3(
+		@InjectConfiguration(PARAM_PID) Configuration configuration)
+		throws Exception {
+
+		Configuration cs = ConfigUtil.getConfigsByServicePid(ca, PARAM_PID);
+
+		assertNull(configuration.getProperties()
```
bjhargrave, line 135: This is much nicer as:

```java
assertThat(Dictionaries.asMap(configuration.getProperties())).doesNotContainKeys("foo", "bar");
```

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
```diff
@@ -0,0 +1,182 @@
...
+	@Test
+	@WithConfiguration(pid = METHOD_PID, properties = {
+		@ConfigEntry(key = "foo", value = "bar")
+	})
+	public void testMethodConfiguration() throws Exception {
+
+		Configuration cs = ConfigUtil.getConfigsByServicePid(ca, METHOD_PID);
+		assertThat(cs).isNotNull();
+		assertEquals("bar", cs.getProperties()
+			.get("foo"));
```
bjhargrave, line 152: This is much nicer as:

```java
assertThat(Dictionaries.asMap(configuration.getProperties())).containsEntry("foo", "bar");
```

One of the reasons I created Dictionaries.asMap was for assertj testing :-)

bjhargrave:
This is making great progress! Well done.

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
kriegfrj: We should have our own custom entry point (soft and hard) for `assertThat(Dictionary)`, as this would make this nicer still. I've noticed it's a common pattern.

Well, I went the extra step:
https://github.com/osgi/osgi-test/pull/174

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/ConfigEntry.java
```diff
@@ -0,0 +1,63 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Inherited
+@Target({
+	ANNOTATION_TYPE
+})
+@Retention(RUNTIME)
+@Documented
+public @interface ConfigEntry {
```
kriegfrj, line 33: Just a thought: I've recently started playing with OSGi MetaTypes. Are we re-inventing the wheel here? Could we leverage the metatype syntax to avoid having to create/maintain our own custom syntax for config entries? Or is there some functionality that metatypes doesn't give us? Or do metatypes give us too much functionality that `ConfigAdmin` doesn't support?

This is just a thought bubble - I don't have a lot of experience with OSGi meta types or with `ConfigAdmin`, so I don't have a strong feel for what is appropriate. I just wanted to throw it out there for consideration.

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationNested.java
kriegfrj: And for tests like this one that have multiple assertions, they'd be even better if we used soft assertions. That one one failure doesn't stop the rest of the assertions from running & giving feedback to the developer.

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/ConfigEntry.java
stbischof: i thought that metatype is is just compiletime to create these metatype xml and not reading an annotation. But i think that only shows that i do not know enough about it.

kriegfrj: You're right - metatype annotations do not have runtime retention and so they couldn't be adapted to this usage.

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
stbischof: done

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
stbischof: done

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
stbischof: done

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
stbischof: done

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationNested.java
stbischof: done

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
```diff
@@ -91,51 +83,51 @@ public void testFieldConfiguration() throws Exception {
...
+	public void testParameterConfiguration(@InjectConfiguration(PARAM_PID) Configuration configuration)
 		throws Exception {
 
 		Configuration cs = ConfigUtil.getConfigsByServicePid(ca, PARAM_PID);
 		assertThat(cs).isEqualTo(configuration);
 
-		assertNull(configuration.getProperties()
-			.get("foo"));
+		assertThat(Dictionaries.asMap(cs.getProperties()))
+			.containsExactlyEntriesOf(Dictionaries.asMap(configuration.getProperties()));
```
bjhargrave, line 93: I just updated DictionaryAssert to include method variants which take Dictionary arguments. So you can now say:

```java
assertThat(cs.getProperties()).containsExactlyInAnyOrderEntriesOf(configuration.getProperties());
```

org.osgi.test.junit5.cm/test.bndrun
```diff
@@ -51,4 +51,6 @@
 	org.osgi.service.cm;version='[1.6.0,1.6.1)',\
 	org.osgi.test.common;version='[0.10.0,0.10.1)',\
 	org.osgi.test.junit5.cm;version='[0.10.0,0.10.1)',\
-	org.osgi.test.junit5.cm-tests;version='[0.10.0,0.10.1)'
+	org.osgi.test.junit5.cm-tests;version='[0.10.0,0.10.1)',\
+	org.osgi.test.assertj.framework;version='[0.10.0,0.10.1)',\
```
bjhargrave, line 55: Please sort.

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationNested.java
```diff
@@ -48,13 +46,23 @@ public void test_class_method1(@InjectConfiguration("class") Configuration c_cla
...
+		softly.assertThat(c_class_Inner1)
+			.isNull();
+		softly.assertThat(c_class_Inner1_Method1)
+			.isNull();
+		softly.assertThat(c_class_Inner1_Method2)
+			.isNull();
+		softly.assertThat(c_class_Inner2)
+			.isNull();
+		softly.assertThat(c_class_Inner2_Method1)
+			.isNull();
```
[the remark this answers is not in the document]
stbischof: thank you!

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
stbischof: i have seen the pr. that helps!

org.osgi.test.junit5.cm/test.bndrun
stbischof: maybe this would be a nice option for bnd

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationNested.java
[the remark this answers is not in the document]
stbischof: Will add the comment.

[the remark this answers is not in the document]
kriegfrj: You can also use the SoftAssertionsExtension which injects a SoftAssertionsProvider and calls assertAll on it automatically.

org.osgi.test.junit5.cm/test.bndrun
kriegfrj: Ask, and you shall receive:
```
-runstartlevel: \
    order=sortbynameversion,\
    begin=-1
```
Add the above to `test.bndrun` (before the -runbundles directive would be best, I think) and you should be good to go!

kriegfrj: I'm not sure why we haven't yet done this for the other bndrun files. Perhaps also a common `.bndrun` somewhere in cnf/includes might be nice so that we can reference this (and other common configuration) in all our bndrun files.

stbischof:
I will make make the BlockingConfigurationHandler injectable that it can be used in the tests

org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/ConfigurationExtension.java
```diff
@@ -0,0 +1,407 @@
...
+					.isAssignableFrom(Dictionary.class)) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	@Override
+	public void beforeAll(ExtensionContext extensionContext) throws Exception {
+		bundleContext = FrameworkUtil.getBundle(extensionContext.getRequiredTestClass())
```
bjhargrave, line 273: You should not get the BundleContext directly. Use BundleContextExtension to get a BundleContext if needed.

org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/ConfigurationExtension.java
```diff
@@ -0,0 +1,407 @@
...
+		}
+	}
+
+	@Override
+	public Object resolveParameter(ParameterContext parameterContext, ExtensionContext extensionContext)
+		throws ParameterResolutionException {
+
+		try {
+
+			BundleContext bundleContext = FrameworkUtil.getBundle(extensionContext.getRequiredTestClass())
```
bjhargrave, line 175: You should not get the BundleContext directly. Use BundleContextExtension to get a BundleContext if needed.

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+	@BeforeAll
+	public static void beforeAll(@InjectService ConfigurationAdmin configurationAdmin) {
+
+		ca = configurationAdmin;
+	}
+
+	@AfterAll
+	public static void afterAll() {
+		BundleContext bundleContext = FrameworkUtil.getBundle(ConfigAnnotationTest.class)
```
bjhargrave, line 57: Use BundleContextExtension in the tests.

stbischof: thx.

org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/ConfigurationExtension.java
stbischof: done

org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/ConfigurationExtension.java
```diff
@@ -0,0 +1,407 @@
...
+		}
+
+		storeConfigCopy(extensionContext);
+		handleAnnotationsOnActiveElement(extensionContext);
+
+	}
+
+	private ConfigurationAdmin getConfigAdmin() {
+
+		return bundleContext.getService(sref);
```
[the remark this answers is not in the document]
stbischof: done

stbischof:
Configuration Admin service has been unregistered before calling afterAll of ConfigurationExtension.
```Configuration Admin service has been unregistered
	at org.apache.felix.cm.impl.ConfigurationAdminImpl.getConfigurationManager(ConfigurationAdminImpl.java:325)
	at org.apache.felix.cm.impl.ConfigurationAdminImpl.listConfigurations(ConfigurationAdminImpl.java:185)
	at org.osgi.test.junit5.cm.ConfigUtil.getAllConfigurations(ConfigUtil.java:85)
	at org.osgi.test.junit5.cm.ConfigUtil.resetConfig(ConfigUtil.java:107)
	at org.osgi.test.junit5.cm.ConfigurationExtension.reset(ConfigurationExtension.java:356)
	at org.osgi.test.junit5.cm.ConfigurationExtension.afterAll(ConfigurationExtension.java:348)
```

bjhargrave:
> @bjhargrave i do not see a way to handle it with ServiceExtention. any ideas?

I am debugging this. We currently have a problem with CloseableBundleContext and its parent child nesting. We are not properly handling service use counting. I have a PR coming to fix.

bjhargrave:
@stbischof I rebased on my fix and added some more commits. Please look them over and squash them when you are happy.

bjhargrave:
@kriegfrj Are you ok to merge this PR?

kriegfrj:
> @kriegfrj Are you ok to merge this PR?

I'll have a detailed look first thing tomorrow, but probably yes.

org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/ConfigurationExtension.java
```diff
@@ -0,0 +1,371 @@
...
+
+	public void updateConfigurationRespectNew(Configuration configurationToBeUpdated,
+		Dictionary<String, Object> newConfigurationProperties, boolean isNewConfiguraion)
+		throws InterruptedException, IOException {
+		if (configurationToBeUpdated != null) {
+			if (newConfigurationProperties != null
+				&& !ConfigUtil.isDictionaryWithNotSetMarker(newConfigurationProperties)) {
+				// has relevant Properties to update
+				blockingConfigHandler.update(configurationToBeUpdated, newConfigurationProperties, 1000);
+			} else if (isNewConfiguraion) {
```
kriegfrj, line 365: `isNewConfiguraion` -> `isNewConfiguration`. We can thank IDEs for the propagation of typos like this... :smile:

kriegfrj (changes requested):
Thanks for all your hard work on this @stbischof and @bjhargrave !

I have found a few minor cleanups which shouldn't take you more than a few minutes to fix.

However, there was one more serious suggestion for an API improvement which I put forward about defining multiple `@Inject` annotations, which would enable us to achieve compile-time sanity checking rather than relying on runtime. Thoughts on this one appreciated.

.github/workflows/cibuild.yml
```diff
@@ -91,3 +91,9 @@ jobs:
       with:
         file: ./org.osgi.test.junit5/target/site/jacoco-it/jacoco.xml
         name: junit5
+    - name: Upload "junit5.cm" coverage to Codecov
+      if: (matrix.codecov)
+      uses: codecov/codecov-action@v1.0.12
```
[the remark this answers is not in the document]
stbischof: updated

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/config/package-info.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright (c) OSGi Alliance (2019). All Rights Reserved.
```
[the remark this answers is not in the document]
stbischof: ok

org.osgi.test.junit5.cm/logback.xml
```diff
@@ -0,0 +1,35 @@
+<!--
+/**
+ * Copyright (c) OSGi Alliance (2019). All Rights Reserved.
```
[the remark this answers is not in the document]
stbischof: ok

org.osgi.test.junit5.cm/pom.xml
```diff
@@ -0,0 +1,255 @@
...
+							<goal>test-jar</goal>
+						</goals>
+						<configuration>
+							<archive>
+								<manifestFile>${project.build.testOutputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+							</archive>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
```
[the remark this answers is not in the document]
stbischof: ok

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationNested.java
```diff
@@ -0,0 +1,173 @@
...
+import org.osgi.service.cm.Configuration;
+import org.osgi.test.assertj.dictionary.DictionaryAssert;
+import org.osgi.test.common.annotation.config.InjectConfiguration;
+import org.osgi.test.common.annotation.config.WithConfiguration;
+import org.osgi.test.common.dictionary.Dictionaries;
+
+@ExtendWith(ConfigurationExtension.class)
+@WithConfiguration(pid = "class")
+@TestMethodOrder(OrderAnnotation.class)
+@Nested
```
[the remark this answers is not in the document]
stbischof: ok

org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/ConfigurationExtension.java
stbischof: ok

org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/ConfigAnnotationNested.java
```diff
@@ -0,0 +1,173 @@
...
+				.isNotNull();
+			softly.assertThat(c_class_Inner2)
+				.isNull();
+			softly.assertThat(c_class_Inner2_Method1)
+				.isNull();
+			softly.assertAll();
+			Dictionary<String, Object> properties = c_class_Inner1.getProperties();
+
+			DictionaryAssert.assertThat(properties)
+				.doesNotContainKey(KEY);
```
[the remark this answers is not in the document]
stbischof: done

stbischof:
> 
> 
> Perhaps it would be better to make it three separate annotations? Perhaps, @InjectConfigurationPID, @InjectConfiguration, @InjectConfigurationFactory? This has a couple of attractive features to it:
> 
>     @InjectConfiguration(withConfig=@WithConfiguration(...)) can simplify to @InjectConfiguration(@WithConfiguration(...))
>     Current implementation allows useless tag definition because all fields are optional
>     Compile-time sanity checking rather than runtime.
> 
> Disadvantage is the increased API footprint. Even this is not a big disadvantage because we have more annotations, but less fields in each one.
> 
> I could go either way on this, but I'm leaning towards separate annotations. I'd be interested in what you guys think.
> 
> Also, there is a typo in the doc - Optional<Condiguration> should be Optional<Configuration>.
> 
> I apologise if this option was already discussed and I missed it.
> 


@bjhargrave I am open for adding annotations. but not sure that is easyer for the user

bjhargrave:
The base case is `@InjectConfiguration("some.pid")`. The `With...Configuration` annotations to are create configurations for a specified pid. Using a `With...Configuration` in the `InjectConfiguration` is just a short hand to avoid having to rewrite the configuration pid string. I don't see the real value in breaking `InjectConfiguration` up into N annotations. The extension can verify the use of `InjectConfiguration` is malformed (no configuration is specified) and fail the test. So I would prefer to leave the `InjectConfiguration` as it currently is (assuming the extension does fail the use of the annotation if no configuration is specified.)

PR #174, opened 2020-08-11
[assertj] Add Dictionary assert

### LINKED ISSUE #173: [assertj] Add assertions for legacy types
OSGi is so old that it has a number of legacy type usages baked deep in its core. Use of the `Dictionary` abstract base class for properties is a classic example.

AssertJ does not have any direct support for some of these legacy classes, as they are not commonly used in modern code that might be using AssertJ.

We already have utility methods like `Dictionaries.asMap()` that allow you to build a "shim" wrapper around the dictionary and view it as a `Map`:
https://github.com/osgi/osgi-test/blob/4bdb06c5e6f7ca2a31ec1f0262477cfc6d851e62/org.osgi.test.common/src/main/java/org/osgi/test/common/dictionary/Dictionaries.java#L44-L54
As @bjhargrave noted (https://github.com/osgi/osgi-test/pull/165/files/8f6ce896a19ca13d2aa1125c7f640ebd77206877#r467947029) one of the reasons he created this was so that the rich set of AssertJ Map assertions would be available for testing dictionaries. This allows:

`assertThat(Dictionaries.asMap(dictionary)).containsKeys("key1", "key2");`

But we could go a step further and create an assertion entry point that can be invoked directly, which would allow this:

`assertThat(dictionary).containsKeys("key1", "key2")`

kriegfrj:
Fixes #173.

### DIFF
--- a/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/dictionary/DictionaryAssert.java
+++ b/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/dictionary/DictionaryAssert.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) OSGi Alliance (2020). All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


kriegfrj:
CI Failure was due to Codecov upload timing out - seems to be an issue with Codecov at the moment.

Issue #121, opened 2020-05-28
[assertj] Integration testing


kriegfrj:
> Maybe we should also be doing full in-framework integration testing of our AssertJ bundle to catch these.

That is probably a necessary thing since the purpose of these is to use for in-framework integration testing. We should be confident that it works well.

_Originally posted by @bjhargrave in https://github.com/osgi/osgi-test/issues/118#issuecomment-632693516_

I just got bitten by this again when I added a package to commons (`org.osgi.test.common.bitmaps`) that I then referenced in assertj - the unit test compiled and ran happily, but then when the junit5 integration test attempted to resolve the assertj bundle it failed because new package wasn't exported. Given that our full integration tests are still pretty fast and the IDE integration is still good, I see no harm with changing the unit tests to run as integration tests which will help to catch issues like these.

PR #176, opened 2020-08-11
Add integration test for DictionarySoftAssertionsProvider 

bjhargrave:


bjhargrave:
I opened https://github.com/joel-costigliola/assertj-core/issues/1964 about the class loader issue in the build.

kriegfrj:
This doesn't really belong in the JUnit 5 bundle as it is not JUnit 5 specific. I think this would be better implemented as part of #121 in the AssertJ bundle itself. I'll see if I can implement this today.

kriegfrj:
> This doesn't really belong in the JUnit 5 bundle as it is not JUnit 5 specific. I think this would be better implemented as part of #121 in the AssertJ bundle itself. I'll see if I can implement this today.

See #178 - I reckon this PR is made obsolete by that one.

PR #178, opened 2020-08-12
Integration tests

### LINKED ISSUE #121: [assertj] Integration testing

kriegfrj:
Change all projects to use integration tests rather than plain JUnit tests.

Because we want to make sure that these classes work properly in OSGi, it is good if all of the tests actually run in an OSGi context. The main disadvantage of this approach is speed, but our tests run pretty fast anyway so the disadvantage is minimal.

Fixes #121.

### DIFF
--- a/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/dictionary/DictionarySoftAssertionsProvider.java
+++ b/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/dictionary/DictionarySoftAssertionsProvider.java
@@ -25,6 +25,8 @@ public interface DictionarySoftAssertionsProvider extends SoftAssertionsProvider
 	 * Create soft assertion for {@link java.util.Dictionary}.
 	 *
 	 * @param actual the actual value.
+	 * @param <K> the type of the keys used in the dictionary
+	 * @param <V> the type of the values used in the dictionary
 	 * @return the created assertion object.
 	 */
 	default <K, V> ProxyableDictionaryAssert<K, V> assertThat(Dictionary<K, V> actual) {
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -65,19 +65,19 @@ jobs:
       if: (matrix.codecov)
       uses: codecov/codecov-action@v1.0.12
       with:
-        file: ./org.osgi.test.assertj.framework/target/site/jacoco/jacoco.xml
+        file: ./org.osgi.test.assertj.framework/target/site/jacoco-it/jacoco.xml
         name: assertj.framework
     - name: Upload "assertj.promise" coverage to Codecov
       if: (matrix.codecov)
       uses: codecov/codecov-action@v1.0.12
       with:
-        file: ./org.osgi.test.assertj.promise/target/site/jacoco/jacoco.xml
+        file: ./org.osgi.test.assertj.promise/target/site/jacoco-it/jacoco.xml
         name: assertj.promise
     - name: Upload "common" coverage to Codecov
       if: (matrix.codecov)
       uses: codecov/codecov-action@v1.0.12
       with:
-        file: ./org.osgi.test.common/target/site/jacoco/jacoco.xml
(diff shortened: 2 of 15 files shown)


org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/dictionary/DictionaryAssert.java
```diff
@@ -45,6 +46,10 @@ protected DictionaryAssert(Map<KEY, VALUE> actual) {
 		return new DictionaryAssert<>(actual);
 	}
 
+	public static <K, V> DictionaryAssert<K, V> assertThat(Hashtable<K, V> actual) {
+		return new DictionaryAssert<>((Dictionary<K, V>) actual);
```
bjhargrave, line 50: Why would you not use MapAssert for Hashtable? Since Dictionaries.asMap will see that Hashtable is-a Map and return it cast to Map. So there is no need for all that since we can just do

```java
public static <K, V> MapAssert<K, V> assertThat(Hashtable<K, V> actual) {
		return new assertThat((Map<K,V>) actual);
}
```

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/dictionary/DictionarySoftAssertionsProvider.java
```diff
@@ -25,11 +26,27 @@
...
+	 * Create soft assertion for {@link java.util.Hashtable}. Provided to allow
+	 * the compiler to resolve the ambiguity for {@code Hashtable} (as it also
+	 * implements {@link java.util.Map}).
+	 *
+	 * @param actual the actual value.
+	 * @param <K> the type of the keys used in the hashtable
+	 * @param <V> the type of the values used in the hashtable
+	 * @return the created assertion object.
+	 */
+	default <K, V> ProxyableDictionaryAssert<K, V> assertThat(Hashtable<K, V> actual) {
```
bjhargrave, line 49: Like above, this should be a MapAssert for Hashtable.

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/dictionary/DictionaryAssert.java
kriegfrj: Because DictionaryAssert has all the same methods that MapAssert does, but also the extra methods that you added which take a Dictionary for comparison. I thought these might be useful. In fact,  we should probably add variants that take a Hashtable to resolve the ambiguity there too.

bjhargrave: At a basic level, I don't think we should do anything for Hashtable. People can cast to Map or DIctionary as they wish. With Maps, people should use Map as the variable type: `Map map = new HashMap();` When using Hashtables, people really ought to declare the variable type as Dictionary (or Map): `Dictionary dict = new Hashtable();`. This then removes all ambiguity.

> but also the extra methods that you added which take a Dictionary for comparison.

Sure, but what about all the people that have maps that want to compare against dictionaries? We have Dictionaries.asDictionary for them. We cannot solve all combinations and I am not sure we should try.

kriegfrj: I am convinced. I'll drop the hashtable patch from this PR.

PR #181, opened 2020-08-12
ServiceExtension improvements

bjhargrave:
We add a public static method to create a managed ServiceConfiguration.
We use a CloseableResource to store the ServiceConfiguration in the
execution context store so that is will be properly closed when the
execution contexts completes. We also use a store keyed to the execution
context as we do in BundleContextExtension. We use the
BundleContextExtension to supply the BundleContext for the
ServiceExtension. Finally, we add support for injecting a service into
a static field.

Also some improvements to BundleContextExtension.

### DIFF
--- a/org.osgi.test.common/src/main/java/org/osgi/test/common/service/ServiceConfigurationKey.java
+++ b/org.osgi.test.common/src/main/java/org/osgi/test/common/service/ServiceConfigurationKey.java
@@ -1,24 +1,25 @@
 package org.osgi.test.common.service;
 
+import static java.util.Objects.requireNonNull;
+
 import java.util.Arrays;
 import java.util.Objects;
 
-import org.osgi.test.common.annotation.InjectService;
+public class ServiceConfigurationKey<S> {
 
-public class ServiceConfigurationKey {
+	final Class<S>	serviceType;
+	final String	filter;
+	final String[]	filterArguments;
+	final int		cardinality;
+	final long		timeout;
 
-	private final int		cardinality;
-	private final String	filter;
-	private final String[]	filterArguments;
-	private final String	serviceName;
-	private final long		timeout;
-
-	public ServiceConfigurationKey(Class<?> serviceType, InjectService injectService) {
-		this.serviceName = serviceType.getName();
-		this.cardinality = injectService.cardinality();
-		this.filter = injectService.filter();
-		this.filterArguments = injectService.filterArguments();
-		this.timeout = injectService.timeout();
+	public ServiceConfigurationKey(Class<S> serviceType, String filter, String[] filterArguments, int cardinality,
+		long timeout) {
+		this.serviceType = requireNonNull(serviceType);
+		this.filter = filter;
+		this.filterArguments = filterArguments;
+		this.cardinality = cardinality;
+		this.timeout = timeout;
 	}
 
 	@Override
@@ -26,7 +27,7 @@ public class ServiceConfigurationKey {
 		final int prime = 31;
 		int result = 1;
 		result = prime * result + Arrays.hashCode(filterArguments);
-		result = prime * result + Objects.hash(cardinality, filter, serviceName, timeout);
+		result = prime * result + Objects.hash(cardinality, filter, serviceType.getName(), timeout);
 		return result;
 	}
 
@@ -38,10 +39,10 @@ public class ServiceConfigurationKey {
 		if (!(obj instanceof ServiceConfigurationKey)) {
 			return false;
 		}
-		ServiceConfigurationKey other = (ServiceConfigurationKey) obj;
+		ServiceConfigurationKey<?> other = (ServiceConfigurationKey<?>) obj;
 		return cardinality == other.cardinality && Objects.equals(filter, other.filter)
-			&& Arrays.equals(filterArguments, other.filterArguments) && Objects.equals(serviceName, other.serviceName)
+			&& Arrays.equals(filterArguments, other.filterArguments)
+			&& Objects.equals(serviceType.getName(), other.serviceType.getName())
 			&& timeout == other.timeout;
 	}
-
 }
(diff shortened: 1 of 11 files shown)


org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/ServiceExtension_ServiceAwareTest.java
```diff
@@ -166,53 +231,60 @@ public void matchByFilter() throws Exception {
...
 
 		@Override
-		ServiceAware<Foo> getServiceAware() {
-			return foo;
+		@Test
+		void test() throws Exception {
+			softly.assertThat(serviceAware.getServiceReference())
+				.as("getServiceReference %s", serviceAware)
+				.isNotNull();
+			softly.assertThat(serviceAware.getServiceReferences())
```
kriegfrj, line 264: I'm curious: What was the reasoning behind moving these here?

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceExtension.java
```diff
@@ -60,16 +63,12 @@
...
-	final static Namespace NAMESPACE = Namespace.create(ServiceExtension.class);
+public class ServiceExtension implements BeforeAllCallback, BeforeEachCallback, ParameterResolver {
 
 	@Override
-	public void beforeEach(ExtensionContext extensionContext) throws Exception {
-		List<Field> fields = findAnnotatedFields(extensionContext.getRequiredTestClass(), InjectService.class);
-
-		BundleContext bundleContext = FrameworkUtil.getBundle(extensionContext.getRequiredTestClass())
-			.getBundleContext();
+	public void beforeAll(ExtensionContext extensionContext) throws Exception {
```
kriegfrj, line 69: Note that for completeness, this should also inject into non-static fields if the `ExtensionContext` has a TestInstance present. This is for if you are running your tests with the `@TestInstance(PER_CLASS)`. There are a couple of situations where this is useful (as I have recently discovered) - if you're using Kotlin (which doesn't support `static`), or if you're using a `@Nested` test (as non-static inner classes cannot have static methods).

`BundleContextExtension` doesn't support this yet either, so unless you're feeling particularly keen then maybe leave this for now and we'll look at a future PR. :smile:

org.osgi.test.common/src/main/java/org/osgi/test/common/service/ServiceConfigurationKey.java
```diff
@@ -1,32 +1,33 @@
...
+	final String	filter;
+	final String[]	filterArguments;
+	final int		cardinality;
+	final long		timeout;
+
+	public ServiceConfigurationKey(Class<S> serviceType, String filter, String[] filterArguments, int cardinality,
+		long timeout) {
+		this.serviceType = requireNonNull(serviceType);
+		this.filter = filter;
+		this.filterArguments = filterArguments;
```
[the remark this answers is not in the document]
stbischof: I think this is done because that  we can use it also as key with the @ServiceSource.

kriegfrj:
Actually, one other thought: is there a reason why you felt it necessary to use `BundleContextExtension` to retrieve the bundle context for `ServiceExtension`? I remember that @rotty3000 discussed it when we first moved to the injection model, and we felt that it was unnecessary because changes made to the `BundleContext` provided by BCE will be reflected in the underlying "live" `BundleContext`. The main reason for using BCE is if you're going to make changes to the bundle context and you want the changes to be automatically unwound; but as the `ServiceExtension` doesn't make changes to the underlying bundle context there doesn't seem to be a need. If you can see a flaw in this reasoning, please let us know.

One argument that I can see for using `BundleContextExtension` is that at some point we might be able to filter the bundle context to hide services that were registered outside of it, to protect your test from the effects of parallelism. But that's a way down the track.

org.osgi.test.common/src/main/java/org/osgi/test/common/service/ServiceConfigurationKey.java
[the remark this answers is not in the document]
bjhargrave: I am not sure it is a great idea. We will create a ServiceConfigurationKey each time we go to the Store to get a ServiceConfiguration so we should make the creation of the key as lightweight as possible. Computing the filter when creating the key seems like wasteful work.

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/ServiceExtension_ServiceAwareTest.java
bjhargrave: The tests ran a test case and then, after the test case completed, examined the ServiceAware provided to the test case. But that violates the lifecycle of the ServiceAware. It should not be valid after the test case completes since we should be clean up resources such as closing the underlying ServiceTracker. This PR fixes the problems with not properly closing the ServiceConfiguration(ServiceAware) and thus the tests had to be modified accordingly.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceExtension.java
bjhargrave: We should fix both in a new PR!

bjhargrave:
> Actually, one other thought: is there a reason why you felt it necessary to use `BundleContextExtension` to retrieve the bundle context for `ServiceExtension`?

I think, as I review the configuration extension PR, that we should drive all bundle context references through our bundle context extension. In the configuration extension PR, we should drive all service references through our service extension. It provides a central management point and ensures all things are properly cleaned up.

> but as the `ServiceExtension` doesn't make changes to the underlying bundle context there doesn't seem to be a need

Of course service extension makes changes to the underlying bundle context. It uses services which are reflected in the underlying bundle context and such "side effects" must be cleaned up between tests so they are not visible in subsequent tests.

org.osgi.test.common/src/main/java/org/osgi/test/common/service/ServiceConfigurationKey.java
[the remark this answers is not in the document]
kriegfrj: My thinking was that to use the key, you'll have to call `hashCode()` and `equals()`, which will involve iterating over the filter template and the filter args independently and calling `equals()` and `hashcode()` on each arg. This applies to every `put()` and `get()` operation. If you use a pre-computed filter, then you only have to compute the hashcode and do the equals on the single string.

But without benchmarking we wouldn't know how much of a difference this will make.

PR #192, opened 2020-08-25
Refactoring and minor changes  - Property Annotation - PropertiesConverter - TargetType

stbischof:
refactoring and minor changes

- Property Annotation
- PropertiesConverter
- TargetType


stbischof:
@juergen-albert

because you gave me the initial idea for @WIthService, could you please have a look an this?

https://raw.githubusercontent.com/osgi/osgi-test/7acb037c9141c32284677a84f81aa0bf4a98f0c2/org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/WithServiceTest.java

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/ServiceProperty.java
```diff
@@ -30,8 +30,9 @@
 })
 @Retention(RUNTIME)
 @Documented
-public @interface ConfigEntry {
+public @interface ServiceProperty {
```
bjhargrave, line 33: Maybe this is just simply `Property`? As in a property per https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module.filtersyntax.

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/ServicePropertyConverter.java
```diff
@@ -0,0 +1,183 @@
...
+			|| entry.type()
+				.equals(Type.PrimitiveArray)) {
+			return array;
+		} else if (entry.type()
+			.equals(Type.Collection)) {
+
+			return Arrays.asList((Object[]) array);
+		}
+
+		return null;
```
bjhargrave, line 110: Returning null here guarantees an NPE in the Hashtable.put call above. If this is really a conversion error, you should probably throw a detailed exception here.

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/ServicePropertyConverter.java
```diff
@@ -0,0 +1,183 @@
...
+
+			case String :
+				if (primitive) {
+					throw new IllegalArgumentException(
+						"@ServiceProperty Could not be Scalar=String and type=primitiveArray at the same time");
+				} else {
+					return new String[length];
+				}
+		}
+		return null;
```
bjhargrave, line 180: Same here.

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/WithService.java
```diff
@@ -0,0 +1,54 @@
...
+@Inherited
+@Target({
+	ElementType.TYPE, ElementType.METHOD
+})
+@Retention(RUNTIME)
+@Documented
+@Repeatable(WithServices.class)
+public @interface WithService {
+
+	String ofMethod() default "";
```
bjhargrave, line 37: We should javadoc all the elements since this is what people will read in their IDEs.

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/WithService.java
```diff
@@ -0,0 +1,54 @@
...
+	 * The services of the Component. If not set Object.class will be used.
+	 *
+	 * @return The services.
+	 */
+	Class<?>[] value() default {
+		Object.class
+	};
+
+	ServiceProperty[] properties() default {
+		@ServiceProperty(key = ServiceProperty.NOT_SET)
```
bjhargrave, line 51: Why isn't the default just an empty array?

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/ServicePropertyConverter.java
```diff
@@ -0,0 +1,183 @@
...
+
+import java.lang.reflect.Array;
+import java.util.Arrays;
+import java.util.Dictionary;
+import java.util.Hashtable;
+
+import org.osgi.test.common.annotation.ServiceProperty.Scalar;
+import org.osgi.test.common.annotation.ServiceProperty.Type;
+
+public class ServicePropertyConverter {
```
bjhargrave, line 26: This would then be named `PropertiesConverter`.

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/WithServiceTest.java
```diff
@@ -0,0 +1,107 @@
...
+import org.junit.jupiter.api.TestInfo;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.osgi.test.common.annotation.InjectService;
+import org.osgi.test.common.annotation.ServiceProperty;
+import org.osgi.test.common.annotation.WithService;
+import org.osgi.test.junit5.service.WithServiceTest.MyFoo;
+import org.osgi.test.junit5.types.Foo;
+
+@ExtendWith(ServiceExtension.class)
+@WithService(ofClass = MyFoo.class, value = Foo.class, properties = {
```
bjhargrave, line 33: The whole concept seems extremely limited. Since you can only use it for classes having a public, no-arg constructor. Or is the some other way to make a service object that you are not showing/testing here? I think the lack of javadoc on WithService is leaving me in the dark.

bjhargrave:
There should be _some_ description of this PR in the opening comment please.

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/WithService.java
stbischof: Then it would generate. Empty hashmap. This will be used to update the config e.g. in configadmin and remove all existing configuration.

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/WithServiceTest.java
stbischof: You cann call a static method inside the testclass to get an object.  Will add javadoc

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/WithService.java
bjhargrave: Well, the code here processes the properties element. So it can decide that an empty array means no service properties, no?

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/WithServiceTest.java
bjhargrave: And tests. You may want to look at org.junit.jupiter.params.provider.MethodSource for inspiration here.

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/WithService.java
stbischof: absolutly agree here.

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/WithServiceTest.java
kriegfrj: Might be nice if it could also call instance methods rather than just static methods - would increase the possible number of use cases. The factory method in this case could, for example, create a service implementation using Mockito and store a reference to it in a test class instance field - then you can perform tests on it using Mockito.verify(). Although, I do wonder if there is much ground gained here over (eg) simply calling bundleContext.registerService() at the start of your test method.

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/ServicePropertyConverter.java
stbischof: ok

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/ServicePropertyConverter.java
stbischof: ok

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/WithService.java
stbischof: sure

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/ServicePropertyConverter.java
stbischof: ok

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/WithServiceTest.java
stbischof: For me the cleanup (unregister) and annotating nestes classes is the interesting part.

bjhargrave: Clean up of registered services is already done by the BundleContextExtension. So you can just use it to get a BundleContext and then use the normal BundleContext.registerService method. So @kriegfrj's comment on just using BundleContext.registerService is quite valid.

kriegfrj: > Will have a look how to get the TestInstance to call more than static methods. in my test i just got null as instance.

`@WithService` tags that are registered at the class level, it would make sense to inject them in `ServiceExtension.beforeAll()`. If your class is running with `TestLifecycle.PER_CLASS`, then the extension context will have a non-null test instance at this stage; otherwise you'd throw an `ExtensionConfigurationException` if you're trying to use a non-static initialiser method (similar to how Jupiter throws an exception if it binds a non-static `@BeforeAll` method on a `PER_CLASS` annotated test class).

For `@WithService` tags registered on an individual test method, you'd inject them in `ServiceExtension.beforeEach()`. The extension context will always have a test class instance available in the beforeEach() callback regardless of the lifecycle mode you're using.
For services register

> For me the cleanup (unregister) and annotating nestes classes is the interesting part.
The unregister is already handled by `BundleContextExtension` if you use `registerService()` through the injected `BundleContext` though. What I wrote above could also be achieved simply by calling `registerService()` in a `@BeforeAll` method. Hence why I see the gain of this annotation as minimal.

If you could elaborate on the "annotating nested classes" part though, perhaps you could enlighten me as to where it starts to get more interesting.

kriegfrj: > `@WithService` tags that are registered at the class level, it would make sense to inject them in `ServiceExtension.beforeAll()`.

This comment applies to nested classes too.

kriegfrj: > Clean up of registered services is already done by the BundleContextExtension. So you can just use it to get a BundleContext and then use the normal BundleContext.registerService method. So @kriegfrj's comment on just using BundleContext.registerService is quite valid.

@stbischof , perhaps given the above comment, we should try and more fully thrash out what the potential gain might be of this annotation before you invest too much effort polishing it up. Some other possible arguments for this approach:

- Declarative configuration (configuration-by-annotation) could be considered "neater" than configuration-in-code
- Configuring the properties dictionary declaratively is somewhat more convenient than building a hashtable in code
- No need to explicitly inject the bundle context. Often you will need the injected BC for other reasons, so you might not save anything here. But there will be cases where you might get away with not injecting the BC at all by using this annotation instead.
- You could allow combining `@WithService` with `@InjectService` functionality, adopting the pattern that `ConfigurationExtension` uses. Makes for some nice symmetry.

On the other hand, I'm not sure that the initialiser method approach gains us much. If you need explicit code to create the service, perhaps you're just as well off to register your service in-code using the injected BundleContext?

At this stage I'm still not 100% sold for or against, though I think having enumerated some of these possible gains I'm leaning towards "for" implementing `@WithService` in some form.

stbischof: I see the point. At the moment i see much more benefits and less written code on the bundleContext side. And you are more flexible if you wand to create new a service of the same class '@beforeEach' Testmethod.

bjhargrave: I think the deal breaker here is that most service instances will need more than a public, no-args constructor to create the service object. So you will always need to run some code to construct the service object. So there is not much to be gained by using an annotation to only handle the service registration if you also must author a service object construction method.

bjhargrave: There is still some value in this PR. I like the rename of ConfigEntry to Property and the related changes. So we can at least keep that.

stbischof: what do you think about TargetType.java?

bjhargrave: > what do you think about TargetType.java?

That also looks useful to keep.

kriegfrj: > I think the deal breaker here is that most service instances will need more than a public, no-args constructor to create the service object. So you will always need to run some code to construct the service object. So there is not much to be gained by using an annotation to only handle the service registration if you also must author a service object construction method.

One special case that I can see is when you'd like to register a mock service implementation. Perhaps a `@WithMockService` implementation that just takes the service type. Then behind the scenes it takes care of calling `mock()` to give you a mock object you can then verify against. Useful for testing client code of simple, stateless services.

bjhargrave: > @WithMockService

Seems rather highly specialized and would imply an implementation coupling to mockito which I would prefer we avoid.

rotty3000: sticking my nose in late but I think I agree that there is no good general solution here. I think publishing services with any of the lifecycle scope methods is really the most practical approach, particularly because you already benefit from cleanup garnered by the BundleContext (you _should_ be using the bundle context supplied by the extension, right?)

bjhargrave: > you _should_ be using the bundle context supplied by the extension, right?)

Yes, you always want to use the BundleContext supplied by BundleContextExtension since it cleans up with proper scoping.

kriegfrj: > > @WithMockService
> 
> Seems rather highly specialized and would imply an implementation coupling to mockito which I would prefer we avoid.

Yes, i would want to avoid tight coupling to Mockito too - at least in the core modules. However,  I'm open to the idea of a mockito module which contained Mockito-specific utilities v such as this. I thi

stbischof:
current Changes:
- rename Property Annotation
- extract and Rename PropertiesConverter
- extract methods into TargetType

org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/PropertiesConverter.java
```diff
@@ -0,0 +1,187 @@
...
+				if (primitive) {
+					return new short[length];
+				} else {
+					return new Short[length];
+				}
+
+			case String :
+				if (primitive) {
+					throw new IllegalArgumentException(
+						"@ServiceProperty Could not be Scalar=String and type=primitiveArray at the same time");
```
bjhargrave, line 178: ServiceProperty is old name

org.osgi.test.junit5.cm/test.bndrun
```diff
@@ -50,6 +50,7 @@
 	junit-platform-launcher;version='[1.6.2,1.6.3)',\
 	org.apache.felix.configadmin;version='[1.9.18,1.9.19)',\
 	org.opentest4j;version='[1.2.0,1.2.1)',\
+	org.osgi.service.cm;version='[1.6.0,1.6.1)',\
```
bjhargrave, line 53: Why is this needed? Felix configadmin already exports the cm package, no?

org.osgi.test.common/src/main/java/org/osgi/test/common/inject/TargetType.java
```diff
@@ -21,29 +22,60 @@ public TargetType(Class<?> type, List<Type> genericType) {
 		this.genericTypes = genericType;
```
bjhargrave, line 22: Should this constructor be private to make sure people use the static `of` methods?

bjhargrave:
@stbischof Lets try and wrap this up so we can make a 0.10.0 release.

bjhargrave:
Updated to address my comments and rebased on main.

bjhargrave (approved):

PR #202, opened 2020-09-11
Parent configuration

bjhargrave:
We now configure the bnd maven plugins in the parent pom using
pluginManagement to set the base configuration for each plugin and
plugins to enable the plugin. We also use executions which are not
inherited by child projects to disable the bnd maven plugins on the
parent project.


kriegfrj:
This will work. But i think the real problem is that our root pom is doing dual duty as a parent pom and as an aggregator pom. This is convenient but not strictly the Maven way. I've seen other Maven projects that put the parent pom in a subdirectory and then the root pom is jblu responsible for aggregating the build of the submodules. That way Maven never actually attempts to build the patent pom project and and this wouldn't happen.

PR #234, opened 2020-11-12
[junit5] Servicefilter respect WildcardType

stbischof:
We should provide a Way to generate an objectClass servicefilter with a wildcard `(objectClass=java.lang.Object)` when using `@InjectService`

Examples that should work:

`@InjectService(filter = "(a=b)") ServiceAware<?> anyServiceParameter`

`@InjectService(filter = "(a=b)", anyObjectClass = true) Object anyServiceParameter`

### DIFF
--- a/org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/InjectService.java
+++ b/org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/InjectService.java
@@ -88,4 +88,17 @@ public @interface InjectService {
 	 */
 	long timeout() default DEFAULT_TIMEOUT;
 
+	/**
+	 * Indicate to take the serviceClass from not from the parameter. Use
+	 * `org.osgi.test.common.annotation.InjectService.AnyService` and the
+	 * ParameterType or FieldType`java.lang.Object` for any Service.
+	 *
+	 * @return The service class.
+	 */
+	Class<?> service() default Object.class;
+
+	public static final class AnyService {
+		private AnyService() {}
+	}
+
 }
--- a/org.osgi.test.common/src/main/java/org/osgi/test/common/service/ServiceConfiguration.java
+++ b/org.osgi.test.common/src/main/java/org/osgi/test/common/service/ServiceConfiguration.java
@@ -33,6 +33,7 @@ import java.util.function.Function;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.Filter;
 import org.osgi.framework.ServiceReference;
+import org.osgi.test.common.annotation.InjectService.AnyService;
 import org.osgi.util.tracker.ServiceTracker;
 import org.osgi.util.tracker.ServiceTrackerCustomizer;
 
@@ -51,10 +52,17 @@ public class ServiceConfiguration<S> implements AutoCloseable, ServiceAware<S> {
 	public ServiceConfiguration(Class<S> serviceType, String format, String[] args, int cardinality, long timeout) {
 		this.serviceType = requireNonNull(serviceType);
 
-		Filter filter = format("(objectClass=%s)", serviceType.getName());
+		Filter filter = null;
 		format = String.format(requireNonNull(format), (Object[]) requireNonNull(args));
 		if (!format.isEmpty()) {
-			filter = format("(&%s%s)", filter.toString(), format);
+
+			if (AnyService.class.equals(serviceType)) {
+				filter = format(format);
+			} else {
+				filter = format("(&(objectClass=%s)%s)", serviceType.getName(), format);
+			}
+		} else {
+			filter = format("(objectClass=%s)", serviceType.getName());
 		}
 		this.filter = filter;
 
--- a/org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceExtension.java
+++ b/org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceExtension.java
@@ -25,6 +25,7 @@ import java.lang.reflect.Modifier;
 import java.lang.reflect.Parameter;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.lang.reflect.WildcardType;
 import java.util.List;
 import java.util.Optional;
 
@@ -131,7 +132,7 @@ public class ServiceExtension implements BeforeAllCallback, BeforeEachCallback, 
 			serviceType = ((ParameterizedType) genericMemberType).getActualTypeArguments()[0];
 		}
 		// The service type must be a raw type
-		if (serviceType instanceof Class) {
+		if (serviceType instanceof Class || serviceType instanceof WildcardType) {
 			return true;
 		}
 		throw new ParameterResolutionException("Can only resolve @" + InjectService.class.getSimpleName()
@@ -171,7 +172,16 @@ public class ServiceExtension implements BeforeAllCallback, BeforeEachCallback, 
 
 		// supportsParameter() If Jupiter does the right thing then this method
 		// should not be called without serviceType being a class
-		assert serviceType instanceof Class;
+		assert serviceType instanceof Class || serviceType instanceof WildcardType;
+
+		if (serviceType instanceof WildcardType) {
+			serviceType = Object.class;
+
+		}
+		if (!injectService.service()
+			.equals(Object.class)) {
+			serviceType = injectService.service();
+		}
 
 		ServiceConfiguration<?> configuration = getServiceConfiguration((Class<?>) serviceType, injectService.filter(),
 			injectService.filterArguments(), injectService.cardinality(), injectService.timeout(), extensionContext);
--- a/org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/AnyServiceTest.java
+++ b/org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/AnyServiceTest.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright (c) OSGi Alliance (2020). All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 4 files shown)


org.osgi.test.common/src/main/java/org/osgi/test/common/annotation/InjectService.java
```diff
@@ -88,4 +88,12 @@
 	 */
 	long timeout() default DEFAULT_TIMEOUT;
 
+	/**
+	 * Indicate require services objectClass filter is not set by the Parameter
+	 * or Field. (objectClass=*).
+	 *
+	 * @return The timeout.
+	 */
+	boolean anyObjectClass() default false;
```
bjhargrave, line 97: Like DS, it would be better to have a `Class<?> service() default Object.class;` element here. See <https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#service.component-anyservice>.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceExtension.java
```diff
@@ -132,7 +133,7 @@ public boolean supportsParameter(ParameterContext parameterContext, ExtensionCon
 			serviceType = ((ParameterizedType) genericMemberType).getActualTypeArguments()[0];
 		}
 		// The service type must be a raw type
-		if (serviceType instanceof Class) {
+		if (serviceType instanceof Class || serviceType instanceof WildcardType) {
```
bjhargrave, line 136: I think we need a dedicated type for this like AnyService in DS. See <https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#org.osgi.service.component.AnyService>.

org.osgi.test.common/src/main/java/org/osgi/test/common/service/ServiceConfiguration.java
```diff
@@ -51,7 +52,8 @@ public ServiceConfiguration(ServiceConfigurationKey<S> key) {
 	public ServiceConfiguration(Class<S> serviceType, String format, String[] args, int cardinality, long timeout) {
 		this.serviceType = requireNonNull(serviceType);
 
-		Filter filter = format("(objectClass=%s)", serviceType.getName());
+		String name= WildcardType.class.equals(serviceType) ? "*": serviceType.getName();
```
bjhargrave, line 55: If we are using AnyService, then we should _not_ put an objectClass expression in the filter. `(objectClass=*)` is just wasteful. It always evaluates to true for a service since all services have an objectClass property.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceExtension.java
bjhargrave: We also need to test that the injection site is type compatible with any possible service object. For example, it is of type Object.

> When using this special service name, the reference field or parameter must be of type java.lang.Object so that any service object can be provided.

bjhargrave (commented):
This should be more inline with the AnyService support coming in DS.

bjhargrave:
> @InjectService(filter = "(a=b)", anyObjectClass = true) Object anyServiceParameter

This should look like

` @InjectService(filter = "(a=b)", service = AnyService.class) Object anyServiceParameter`

stbischof:
does that mean we have to wait until AnyService.class exists, or could i create a `org.osgi.test.common.annotation.AnyService.class

bjhargrave:
> does that mean we have to wait until AnyService.class exists, or could i create a `org.osgi.test.common.annotation.AnyService.class

No, I don't mean that. I would create an inner static final class of InjectService.

```java
public static final class AnyService {
  private AnyService();
}
```

bjhargrave:
You need to fix the merge conflict before it can be reviewed.

stbischof:
Thank, I did not noticed that.

org.osgi.test.common/src/main/java/org/osgi/test/common/service/ServiceConfiguration.java
```diff
@@ -51,10 +52,18 @@ public ServiceConfiguration(ServiceConfigurationKey<S> key) {
...
+		Filter filter = null;
 		format = String.format(requireNonNull(format), (Object[]) requireNonNull(args));
 		if (!format.isEmpty()) {
-			filter = format("(&%s%s)", filter.toString(), format);
+
+			if (AnyService.class.equals(serviceType)) {
+				filter = format(format);
+			} else {
+				filter = format("(objectClass=%s)", serviceType.getName());
+				filter = format("(&%s%s)", filter.toString(), format);
```
bjhargrave, line 63: This seems pointless to format a string and make it into a filter only to then toString the filter! Can't we just do one format call?

```java
filter = format("(&(objectClass=%s)%s)", serviceType.getName(), format);
```

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/WildcardServiceTest.java
```diff
@@ -0,0 +1,98 @@
...
+import org.osgi.test.common.dictionary.Dictionaries;
+import org.osgi.test.common.service.ServiceAware;
+import org.osgi.test.junit5.context.BundleContextExtension;
+
+@ExtendWith(BundleContextExtension.class)
+@ExtendWith(ServiceExtension.class)
+public class WildcardServiceTest {
+
+	@InjectService(filter = "(a=b)")
+	ServiceAware<?> anyServiceField;
```
bjhargrave, line 37: This should look for `(objectClass=java.lang.Object)`. It does not use `service=AnyService.class` which is necessary to get _any_ service. We should not allow generic wildcards to equal AnyService.

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/WildcardServiceTest.java
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	@Test
+	public void testWithServiceAwareWildcardField() throws Exception {
+		assertThat(anyServiceField).isNotNull();
+	}
+
+	@Test
+	public void testWithServiceAwareWildcardParameterImplizit(
+		@InjectService(filter = "(a=b)") ServiceAware<?> anyServiceParameter) throws Exception {
```
bjhargrave, line 51: Same here. Need use to AnyService.

bjhargrave (changes requested):

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/WildcardServiceTest.java
stbischof: That was intended.  WildcardType was for me kind of the same like `AnyType`.

org.osgi.test.common/src/main/java/org/osgi/test/common/service/ServiceConfiguration.java
stbischof: thank you for the hint

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/WildcardServiceTest.java
stbischof: changed the behavior

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/WildcardServiceTest.java
stbischof: changed the behavior

bjhargrave:
This needs to be rebased on the tip of main and conflicts resolved.

stbischof:
rebased. Thx

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/WildcardServiceTest.java
```diff
@@ -0,0 +1,133 @@
...
+	public void test_param_ServiceAwareWildcard_Default(
+		@InjectService(filter = "(a=b)") ServiceAware<?> param_ServiceAwareWildcard_Default) throws Exception {
+		assertThat(param_ServiceAwareWildcard_Default.getServices()).hasSize(1)
+			.hasOnlyElementsOfTypes(Foo.class);
+	}
+
+	public void test_param_ServiceAwareWildcard_AnyService(
+		@InjectService(filter = "(a=b)", service = AnyService.class) ServiceAware<?> param_ServiceAwareWildcard_AnyService)
+		throws Exception {
+		assertThat(field_ServiceAwareWildcard_AnyService.getServices()).hasSize(2)
```
bjhargrave, line 90: You need to assert on `param_ServiceAwareWildcard_AnyService`. Not the field.

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/WildcardServiceTest.java
```diff
@@ -0,0 +1,133 @@
...
+import org.osgi.test.common.annotation.InjectService;
+import org.osgi.test.common.annotation.InjectService.AnyService;
+import org.osgi.test.common.dictionary.Dictionaries;
+import org.osgi.test.common.service.ServiceAware;
+import org.osgi.test.junit5.context.BundleContextExtension;
+import org.osgi.test.junit5.types.Foo;
+
+@ExtendWith(BundleContextExtension.class)
+@ExtendWith(ServiceExtension.class)
+public class WildcardServiceTest {
```
bjhargrave, line 35: Can we call this AnyServiceTest?

bjhargrave (changes requested):
Looks very nice.

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/WildcardServiceTest.java
stbischof: https://github.com/osgi/osgi-test/blob/11102f99746ab842ed4e7971abf911324320ed36/org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/AnyServiceTest.java
done

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/WildcardServiceTest.java
stbischof: oh , thank you. done

bjhargrave (approved):

bjhargrave:
Thanks!

PR #246, opened 2020-11-22
Same toString methodName in all bitmaps

stbischof:
`BundleEventType.class` and `BundleSate.class` uses `toString(int type)` to get a readable name.

`ServiceEventType.class` and `FrameworkEventType.class` have `typeToString(int type)`

we should take the same name for all.


kriegfrj:
Thanks Stefan - please rebase onto main and re-push. I want to make sure that this fixes the the CI error before merging (I think it should).

stbischof:
@kriegfrj Done!

PR #245, opened 2020-11-22
Observe Service-, Framework- and BundleEvents

stbischof:
`Events.class, BundleEvents.class, ServiceEvents.class, FrameworkEvents.class` provides some Predicats to test `Service-, Framework-  and BundleEvents`.

`Events.recordEvents` provides a mechanism to waitFor one or multiple Events that matches a Predicate. It generates a EventRecording stat stores all events.

`EventRecordingAssert` uses this mechanism to record events when executing an action.
'ServiceEventCondition' provides some `assertj.Conditions` (Predicates with an Test-Description)

```
      void exampleIntegrationTest() throws Exception {

        // Setup assert
        EventRecordingAssert eAssert = EventRecordingAssert.assertThatServiceEvent(() -> {
            ServiceRegistration<A> reg = null;
            reg = bc.registerService(A.class, new A() {}, dictionaryOf(k1, v1));
            reg.setProperties(dictionaryOf(k1, v1, k2, v2));
            reg.unregister();
        }, ServiceEvents.isTypeUnregistering(A.class), 20);

        // check whether the Predicate matches or the timeout
        eAssert.isNotTimedOut();

        // get ListAsserts and check them
        eAssert.events()
            .isNotEmpty();

        eAssert.frameworkEvents()
            .isEmpty();

        eAssert.bundleEvents()
            .isEmpty();

        // ListAsserts in combination with Conditions
        eAssert.serviceEvents()
            .areAtLeast(1, isTypeModified())
            .are(hasObjectClass(A.class))
            .areExactly(3, hasObjectClass(A.class))
            .areNot(isTypeModifiedEndmatch())
            .first()
            .isOfType(REGISTERED);

        eAssert.serviceEvents()
            .isNotEmpty()
            .hasSize(3);

        eAssert.serviceEvents()
            .element(0)
            .is(isTypeRegistered())
            .is(hasObjectClass(A.class))
            .is(isTypeRegistered(A.class))
            .is(containServiceProperty(k1, v1))
            .is(matches(REGISTERED, A.class, dictionaryOf(k1, v1)));

        eAssert.serviceEvents()
            .element(1)// ServiceEventAssert
            .is(isTypeModified(A.class))
            .is(containsServiceProperties(dictionaryOf(k1, v1, k2, v2)))
            .serviceReference()// ServiceReferenceAssert
            .serviceProperties()// DictionaryAssert
            .containsEntry(k1, v1);

        eAssert.serviceEvents()
            .element(2)
[...]

PR #253, opened 2020-12-11
ServiceReferenceAssert 

stbischof:
ServiceReferenceAssert

@kriegfrj could you please review? It came from your Comment https://github.com/osgi/osgi-test/pull/245#discussion_r540613956

### DIFF
--- a/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
+++ b/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright (c) OSGi Alliance (2020, 2021). All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.osgi.test.assertj.servicereference;
+
+import static org.osgi.test.assertj.bundle.BundleAssert.BUNDLE;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+
+import org.assertj.core.api.AbstractComparableAssert;
+import org.assertj.core.api.InstanceOfAssertFactories;
+import org.assertj.core.api.InstanceOfAssertFactory;
+import org.assertj.core.api.ListAssert;
+import org.osgi.framework.Bundle;
+import org.osgi.framework.FrameworkUtil;
+import org.osgi.framework.ServiceReference;
+import org.osgi.framework.Version;
+import org.osgi.test.assertj.bundle.BundleAssert;
+import org.osgi.test.assertj.dictionary.DictionaryAssert;
+import org.osgi.test.common.dictionary.Dictionaries;
+
+public abstract class AbstractServiceReferenceAssert<SELF extends AbstractServiceReferenceAssert<SELF, ACTUAL>, ACTUAL extends ServiceReference<?>>
+	extends AbstractComparableAssert<SELF, ACTUAL> {
+
+	protected AbstractServiceReferenceAssert(ACTUAL actual, Class<?> selfType) {
+		super(actual, selfType);
+	}
+
+	public DictionaryAssert<String, Object> hasServicePropertiesThat() {
+		return new DictionaryAssert<String, Object>(Dictionaries.asDictionary(actual));
+	}
+
+	public BundleAssert isRegisteredInBundleThat() {
+		return isNotNull().extracting(actual -> actual.getBundle(), BUNDLE)
+			.as(actual + ".bundle");
+	}
+
+	@SuppressWarnings("rawtypes")
+	static InstanceOfAssertFactory<List, ListAssert<Bundle>> BUNDLE_LIST = InstanceOfAssertFactories.list(Bundle.class);
+
+	public ListAssert<Bundle> hasUsingBundlesThat() {
+		return isNotNull().extracting(bundleContext -> Arrays.asList(bundleContext.getUsingBundles()), BUNDLE_LIST)
+			.as(actual + ".usingBundles");
+	}
+
+	public SELF isRegisteredInBundle(Bundle bundle) {
+
+		isNotNull();
+		if (!Objects.equals(actual.getBundle(), bundle)) {
+			throw failure("%nExpecting%n <%s>%nto be registered in Bundle:%n <%s>%n but it was not", actual, bundle);
+		}
+
+		return myself;
+	}
+
+	public SELF isRegisteredInBundle(String bundleSymbolicName) {
+
+		isNotNull();
+		if (actual.getBundle() == null || !actual.getBundle()
+			.getSymbolicName()
+			.equals(bundleSymbolicName)) {
+			throw failure("%nExpecting%n <%s>%nto be registered in Bundle with SymbolicName:%n <%s>%n but it was not",
+				actual, bundleSymbolicName);
+		}
+
+		return myself;
+	}
+
+	public SELF isRegisteredInBundle(String bundleSymbolicName, String version) {
+
+		isNotNull();
+		Bundle acBundle = actual.getBundle();
+		String bsnva = acBundle == null ? "no bundle registered"
+			: acBundle.getSymbolicName() + ":" + acBundle.getVersion();
+		String bsnvExp = bundleSymbolicName + ":" + version;
+		if (acBundle == null || !(acBundle.getSymbolicName()
+			.equals(bundleSymbolicName)
+			&& acBundle.getVersion()
+				.equals(Version.parseVersion(version)))) {
+			throw failureWithActualExpected(bsnva, bsnvExp,
+				"%nExpecting%n <%s>%nto be registered in Bundle with SymbolicName and Version: <%s>%n but it was %s",
+				actual, bsnvExp, bsnva);
+		}
+		return myself;
+	}
+
+	public SELF isAssignableTo(Class<?> clazz) {
+
+		return isAssignableTo(FrameworkUtil.getBundle(clazz), clazz.getName());
+	}
+
+	public SELF isAssignableTo(Bundle bundle, String className) {
+
+		isNotNull();
+		if (!actual.isAssignableTo(bundle, className)) {
+			throw failure("%nExpecting%n <%s>%nto be assignable to:%n <%s>,<%s>%n but it was not", actual, bundle,
+				className);
+		}
+		return myself;
+	}
+
+	public SELF isBeingUsedBy(Bundle... bundles) {
+
+		as("isBeingUsedBy").hasUsingBundlesThat()
+			.contains(bundles);
+		return myself;
+	}
+
+	public SELF isBeingUsedBy(String... bundleSymbolicNames) {
+
+		for (String bsn : bundleSymbolicNames) {
+			as("isBeingUsedBy").hasUsingBundlesThat()
+				.anyMatch((b) -> bsn.equals(b.getSymbolicName()));
+		}
+		return myself;
+	}
+
+}
--- a/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/ServiceReferenceAssert.java
+++ b/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/ServiceReferenceAssert.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) OSGi Alliance (2020, 2021). All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.osgi.test.assertj.servicereference;
+
+import org.osgi.framework.ServiceReference;
+
+public class ServiceReferenceAssert
+	extends AbstractServiceReferenceAssert<ServiceReferenceAssert, ServiceReference<?>> {
+
+	public ServiceReferenceAssert(ServiceReference<?> actual) {
+		super(actual, ServiceReferenceAssert.class);
+	}
+
+	public static ServiceReferenceAssert assertThat(ServiceReference<?> actual) {
+		return new ServiceReferenceAssert(actual);
+	}
+}
--- a/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/ServiceReferenceSoftAssertions.java
+++ b/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/ServiceReferenceSoftAssertions.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright (c) OSGi Alliance (2020, 2021). All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.osgi.test.assertj.servicereference;
+
+import org.assertj.core.api.SoftAssertions;
+import org.osgi.framework.ServiceReference;
+
+/**
+ * Soft assertions for {@link ServiceReference}s.
+ */
+public class ServiceReferenceSoftAssertions extends SoftAssertions implements ServiceReferenceSoftAssertionsProvider {}
--- a/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/ServiceReferenceSoftAssertionsProvider.java
+++ b/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/ServiceReferenceSoftAssertionsProvider.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) OSGi Alliance (2020, 2021). All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.osgi.test.assertj.servicereference;
+
+import org.assertj.core.api.SoftAssertionsProvider;
+import org.osgi.framework.ServiceReference;
+
+public interface ServiceReferenceSoftAssertionsProvider extends SoftAssertionsProvider {
+	/**
+	 * Create assertion for {@link org.osgi.framework.ServiceReference}.
+	 *
+	 * @param actual the actual value.
+	 * @return the created assertion object.
+	 */
+	default ServiceReferenceAssert assertThat(ServiceReference<?> actual) {
+		return proxy(ServiceReferenceAssert.class, ServiceReference.class, actual);
+	}
+}
--- a/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/package-info.java
+++ b/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/package-info.java
@@ -1,0 +1,19 @@
+/*
+ * Copyright (c) OSGi Alliance (2020, 2021). All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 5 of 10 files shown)


org.osgi.test.common/src/main/java/org/osgi/test/common/dictionary/Dictionaries.java
```diff
@@ -441,6 +443,83 @@ public String toString() {
...
+		}
+
+		@Override
+		public Object remove(Object key) {
+			throw new UnsupportedOperationException();
+		}
+	}
+
+	public static <K, V> Dictionary<String, Object> asDictionary(ServiceReference<?> serviceReference) {
+		return new ServiceReferenceAsDictionary(serviceReference);
```
bjhargrave, line 520: Why do we need this? ServiceReference has a getProperties method which returns a dictionary which can then be asserted on. https://docs.osgi.org/javadoc/osgi.core/7.0.0/org/osgi/framework/ServiceReference.html#getProperties--

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
```diff
@@ -0,0 +1,118 @@
...
+import org.assertj.core.api.InstanceOfAssertFactory;
+import org.assertj.core.api.ListAssert;
+import org.osgi.framework.Bundle;
+import org.osgi.framework.FrameworkUtil;
+import org.osgi.framework.ServiceReference;
+import org.osgi.test.assertj.bundle.BundleAssert;
+import org.osgi.test.assertj.dictionary.DictionaryAssert;
+import org.osgi.test.common.dictionary.Dictionaries;
+
+public abstract class AbstractServiceReferenceAssert<SELF extends AbstractServiceReferenceAssert<SELF, ACTUAL>, ACTUAL extends ServiceReference<?>>
```
bjhargrave, line 35: I am not sure why this is necessary. Are you testing framework implementation to see if they properly implement the OSGi spec? Otherwise, why not just assume the framework is properly implemented. The support we have OSGi testing should be focused on testing bundles people write. To the extent they are useful for testing OSGi framework specification implementation, that is great, but we should not worry too much about that.

org.osgi.test.common/src/main/java/org/osgi/test/common/dictionary/Dictionaries.java
stbischof: Oh, this came with R7. `osgi-test` uses R6 for core. Could we update to R7?

The goal is to get a fast and fluent way to came from `ServiceEventAssert` over `ServiceReferenceAssert` to `DictionaryAssert` via `has...That()`

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
stbischof: As described above the main goal it to have a fluent was to came to DictionaryAssert.
The other Methods are wishes from @kriegfrj  https://github.com/osgi/osgi-test/pull/245#discussion_r540613956

org.osgi.test.common/src/main/java/org/osgi/test/common/dictionary/Dictionaries.java
kriegfrj: Agreed, it would be simpler to use ServiceReference.getProperties(), but as @stbischof pointed out it's not available until R7.

bjhargrave: So perhaps we update the osgi.core dependency to 7.0.0? @rotty3000 ?

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
bjhargrave: But if we update to osgi.core:7..0.0, then you just need to do a dictionary assert on ServiceReference.getProperties and then you don't need this class.

bjhargrave: > is think that a ServiceReferenceAssert is very usefull for a lot of developers.
> will add more tests like `hasObjectClass`

I think you are just testing the framework impl here. Do you think a bundle developer needs to test that his service is properly registered?

stbischof: Sure, you do not need this for a single with DS- registered Component. 
And I will never use more then the bridge to 
```
 `DictionaryAssert` via `hasServicePropertiesThat`
 `BundleAssert` via `hasBundleThat`
``` 
But @kriegfrj  asked me to implement all test when i touch this `ServiceReferenceAssert`.

I had a look on the Tests of DataInMotion and talked to them. They used such tests and checks change-counts and tests services in this detail. And they use this in there OSGi-Spec implementations AND normal bundles. This happened most times when observing ServiceEvents ( like witha mecanism like https://github.com/osgi/osgi-test/pull/257) and then asserting the Reference of a cached Event.

stbischof: > But if we update to osgi.core:7..0.0, then you just need to do a dictionary assert on ServiceReference.getProperties and then you don't need this class.

sure. you can to it by your own. I just like the fluent way.

juergen-albert: I think this is very useful. We e.g. have a Component that can receive a configuration a few other components  and registers a couple service. When we test it, we check if the services are around, have the right properties and that they are bound to each other correctly, because modify e.g. the reference targets via the configuration.

kriegfrj: > I am not sure why this is necessary. Are you testing framework implementation to see if they properly implement the OSGi spec? Otherwise, why not just assume the framework is properly implemented. The support we have OSGi testing should be focused on testing bundles people write. To the extent they are useful for testing OSGi framework specification implementation, that is great, but we should not worry too much about that.

I sympathise with this, which is why I hadn't gone to the effort of writing `ServiceReferenceAssert` myself - I thought it would be rarely used (mainly in framework compliance testing scenarios, which is useful though not our top priority). And if you're using DS, then you won't often touch a `ServiceReference` directly in your code anyway. That said, there are still times when you need to directly access `ServiceReference` and I can see that this assertion might be useful in those cases too.

Add to that, if someone else has identified a need in their own code and was keen to do the work, I thought "why not let them"?

The only concern I have that, if it is to be done, it ought to be done as properly and completely as is practical. It's typically not too much effort to implement most of the useful assertion methods for an assert class once you've already committed to the new assertion class itself. And it can be annoying to think that you have the assertion available only to find it's missing the method you actually needed.

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
```diff
@@ -0,0 +1,117 @@
...
+	extends AbstractComparableAssert<SELF, ACTUAL> {
+
+	protected AbstractServiceReferenceAssert(ACTUAL actual, Class<?> selfType) {
+		super(actual, selfType);
+	}
+
+	public DictionaryAssert<String, Object> hasServicePropertiesThat() {
+		return new DictionaryAssert<String, Object>(actual.getProperties());
+	}
```
kriegfrj, line 43: This depends on us moving the baseline OSGi support to R7 and we're not 100% sure if we're going down that yet - I think it would be courtesy to wait to hear from @rotty3000 for his opinion as I know he's using osgi-test and he may require R6 compatibility.

Leave it this way for now, but make sure you don't throw away the `ServiceReferenceAsDictionary` implementation in case we need to roll back to R6.

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
```diff
@@ -0,0 +1,117 @@
...
+		return isNotNull().extracting(bundleContext -> Arrays.asList(bundleContext.getUsingBundles()), BUNDLE_LIST)
+			.as(actual + ".usingBundles");
+	}
+
+	public SELF isRegisteredInBundle(Bundle bundle) {
+
+		isNotNull();
+		BundleAssert.assertThat(actual.getBundle())
+			.as("isRegisteredInBundle %s:%s", bundle)
+			.isEqualTo(bundle);
```
kriegfrj, line 63: For assertions I find it's usually better not to delegate to another assertion syntax using the "as" formula. You can produce more fluent error messages by calling `failure()` directly (as you've done in with `isAssignableTo()`).

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
```diff
@@ -0,0 +1,117 @@
...
+		return myself;
+	}
+
+	public SELF isRegisteredInBundle(String bundleSymbolicName) {
+
+		isNotNull();
+		BundleAssert.assertThat(actual.getBundle())
+			.as("isRegisteredInBundle %s", bundleSymbolicName)
+			.hasSymbolicName(bundleSymbolicName);
```
kriegfrj, line 72: For assertions I find it's usually better not to delegate to another assertion syntax using the "as" formula. You can produce more fluent error messages by calling `failure()` directly (as you've done in with `isAssignableTo()`).

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
```diff
@@ -0,0 +1,117 @@
...
+		return myself;
+	}
+
+	public SELF isRegisteredInBundle(String bundleSymbolicName, String version) {
+
+		isNotNull();
+		BundleAssert.assertThat(actual.getBundle())
+			.as("isRegisteredInBundle %s:%s", bundleSymbolicName, version)
+			.hasSymbolicName(bundleSymbolicName)
+			.hasVersion(version);
```
kriegfrj, line 82: For assertions I find it's usually better not to delegate to another assertion syntax using the "as" formula. You can produce more fluent error messages by calling `failure()` directly (as you've done in with `isAssignableTo()`).

For this one, I would concatenate the bsn and version into a string and compare those. You can then use these strings as your "actual" and "expected" parameters for `failureWithActualExpected()` to improve the GUI experience.

kriegfrj (changes requested):
Looking good.

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
stbischof: https://gist.github.com/stbischof/943a8ed64470a385d8224313289568ad

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
stbischof: done. thx for the hint

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
rotty3000: For now I would rather keep it R6 compatible.

bjhargrave: OK, but then we should not make API which we won't need in the future when we move to R7. So we need a method to make a Dictionary from a ServiceReference. Then one can assert on that Dictionary. When we move to R7, the implementation of the method to make a Dictionary from a ServiceReference changes.

kriegfrj: Agreed. @stbischof - looks like you'll need to dust off that `ServiceReferenceAsDictionary` implementation and slot it in here for now. :smile:

stbischof: i added 
```
public static <K, V> Dictionary<String, Object> asDictionary(ServiceReference<?> serviceReference) {
		return new ServiceReferenceAsDictionary(serviceReference);
		
```

bjhargrave (commented):
Looks good. Still needs some test for the soft assertions.

stbischof:
@bjhargrave  I added a Test but I am not sure what you really expect. Could you please give me an Example?

bjhargrave:
> Could you please give me an Example?

You can see some test methods in https://github.com/osgi/osgi-test/blob/main/org.osgi.test.assertj.framework/src/test/java/org/osgi/test/assertj/dictionary/DictionaryAssertTest.java are using the DictionarySoftAssertions assertion type.

Also, in addition to testing that expected results _do not_ throw an assertion failure, you also need to test that unexpected results _do_ throw an assertion failure. See 

https://github.com/osgi/osgi-test/blob/9752c6b3fa53f7554a7ea0d1f940572200d7787b/org.osgi.test.assertj.framework/src/test/java/org/osgi/test/assertj/dictionary/DictionaryAssertTest.java#L198-L203 

were we also test that unexpected results throw an assertion error.

bjhargrave (approved):

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
```diff
@@ -0,0 +1,133 @@
...
+
+	protected AbstractServiceReferenceAssert(ACTUAL actual, Class<?> selfType) {
+		super(actual, selfType);
+	}
+
+	public DictionaryAssert<String, Object> hasServicePropertiesThat() {
+		return new DictionaryAssert<String, Object>(Dictionaries.asDictionary(actual));
+	}
+
+	public BundleAssert hasBundleThat() {
```
kriegfrj, line 48: I think this would read better as `isRegisteredInBundleThat()` and ties in better with the other assertion methods.

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/ServiceReferenceAssert.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (c) OSGi Alliance (2020). All Rights Reserved.
+ *
```
kriegfrj, line 3: Add 2021

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/ServiceReferenceSoftAssertions.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright (c) OSGi Alliance (2020). All Rights Reserved.
```
kriegfrj, line 2: Add 2021

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/ServiceReferenceSoftAssertionsProvider.java
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright (c) OSGi Alliance (2020). All Rights Reserved.
```
kriegfrj, line 2: Add 2021

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/package-info.java
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright (c) OSGi Alliance (2020). All Rights Reserved.
```
kriegfrj, line 2: Add 2021

org.osgi.test.common/src/test/java/org/osgi/test/common/dictionary/DictionariesTestCase.java
```diff
@@ -67,10 +80,24 @@ public void setUp() throws Exception {
 		map.put("key4", "value4");
 		map.put("key5", "value5");
 
+		sr = mock(ServiceReference.class);
```
kriegfrj, line 83: I'm a big fan of Mockito, but its use does come at a price. Given that you're not actually using any of Mockito's `verify()` methods, I suspect in this case it might have been easier (and more readable/maintainable in the longer term) to directly create in source a "fake" implementation of `ServiceReference`, rather than dynamically create one using `mock()` and `when()`. I'd leave it as is for now though, more of a note for future reference.

org.osgi.test.assertj.framework/src/test/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssertTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+		ServiceReference<?> sr = mock(ServiceReference.class);
+		setActual(sr);
+
+		when(sr.isAssignableTo(null, null)).thenReturn(true);
+		assertPassing("isNot", x -> aut.isAssignableTo(null, null), null);
+
+		when(sr.isAssignableTo(null, null)).thenReturn(false);
+		assertFailing("is", x -> aut.isAssignableTo(null, null), null)
+			.hasMessageMatching("(?si).*to be assignable to:.*but it was not.*");
```
kriegfrj, line 30: Better to pass in more sensible values here and then check that the message contains those values. At the moment we might be putting the wrong information into the error message (or omitting it) and we wouldn't know. Additional advantage is that you don't have to change the mock method's return value - just pass in different values.

org.osgi.test.common/src/main/java/org/osgi/test/common/dictionary/Dictionaries.java
kriegfrj: For posterity: it was resolved to keep R6 compatibility so `ServiceReferenceAsDictionary` was retained.

kriegfrj (changes requested):
* One typo that needs fixing
* Update copyright year
* Suggested method rename for `hasBundleThat()`.

org.osgi.test.common/src/test/java/org/osgi/test/common/dictionary/DictionariesTestCase.java
stbischof: In my first approach i had a `FakeServiceReference implements ServiceReference`. But that indicated bnd to add a
`Import-Package: org. osgi.framework;version="[1.8,1.9) ` but our eclipse hast an framework package version is `>1.9`

org.osgi.test.assertj.framework/src/test/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssertTest.java
stbischof: done

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
```diff
@@ -0,0 +1,133 @@
...
+		return myself;
+	}
+
+	public SELF isRegisteredInBundle(String bundleSymbolicName) {
+
+		isNotNull();
+		if (actual.getBundle() == null || !actual.getBundle()
+			.getSymbolicName()
+			.equals(bundleSymbolicName)) {
+			throw failure("%nExpecting%n <%s>%nto be registered in Bundle with SymbilicName:%n <%s>%n but it was not",
```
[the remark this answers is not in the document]
stbischof: updated

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
```diff
@@ -0,0 +1,133 @@
+/*
+ * Copyright (c) OSGi Alliance (2020). All Rights Reserved.
```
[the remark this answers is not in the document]
stbischof: updated

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/ServiceReferenceAssert.java
stbischof: updated

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/ServiceReferenceSoftAssertions.java
stbischof: updated

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/ServiceReferenceSoftAssertionsProvider.java
stbischof: updated

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/package-info.java
stbischof: updated

org.osgi.test.common/src/test/java/org/osgi/test/common/dictionary/DictionariesTestCase.java
kriegfrj: You could override the Import-Package to make it `[1.8,2.0)`. I'm actually surprised that Bnd didn't do this by default?

org.osgi.test.common/src/test/java/org/osgi/test/common/dictionary/DictionariesTestCase.java
```diff
@@ -225,6 +276,7 @@ public void null_checks() {
 		assertThat(asMap.remove(null)).isNull();
 		assertThat(asDictionary.get(null)).isNull();
 		assertThat(asDictionary.remove(null)).isNull();
+		assertThat(asDictionarySr.get(null)).isNull();
```
kriegfrj, line 279: When doing multiple assertions in a test method, it is usually helpful to use the `as()` method of AssertJ to describe the assertion. That way when it fails you've got more of an idea what's going on rather than having to dig through the source (and sometimes the line numbers change when you edit). Eg:

```java
assertThat(asMap.remove(null)).as("asMap.remove").isNull();
assertThat(asDictionary.get(null)).as("asDictionary.get").isNull();
```

It can be a bit laborious though so in simple tests I am usually a bit more lenient with this rule. Just letting you know.

kriegfrj:
Good work!

PR #256, opened 2020-12-15
[assertj] Conditions to test Objects and Lists

stbischof:
AssertJ provides an way to Extend Asserations [see AssertJ](https://assertj.github.io/doc/#assertj-core-conditions)

Conditions helpes to test Objects and Lists using is/has, isNot/hasNot, contains, doesNotContain ...
They are kind of a described Predicate.

This PR provides a these Conditions.

Would be nice to get some FeedbackThe
The output-message of join-Conditions will look better in next assertj version. [fix in assertj](https://github.com/assertj/assertj-core/pull/2074)

```java
       List<ServiceEvents> serviceEvents;
    
        // ListAsserts in combination with Conditions
        serviceEvents
            .areAtLeast(1, typeModified())
            .are(hasObjectClass(A.class))
            .areExactly(3, objectClass(A.class))
            .areNot(typeModifiedEndmatch())

           serviceEvents
            .element(0)
            .is(typeRegistered())
            .is(objectClass(A.class))
            .is(typeRegistered(A.class))
            .is(containServiceProperty(k1, v1))
            .is(matches(REGISTERED, A.class, dictionaryOf(k1, v1)));
```

PR #257, opened 2020-12-16
EventRecording and EventRecordingAssert to test events while an action

stbischof:
`MonitoringAssertion.executeAndObserve()` provides a mechanism to waitFor one or multiple Events that matches a Predicate. 

```
      void exampleIntegrationTest() throws Exception {
		MonitoringAssertion.executeAndObserve(() -> {

			ServiceRegistration<Serializable> sr = bc.registerService(Serializable.class, new Serializable() {},
				Dictionaries.dictionaryOf("k", "v0"));

			sr.setProperties(Dictionaries.dictionaryOf("k", "v1"));
			sr.setProperties(Dictionaries.dictionaryOf("k", "v2"));
			sr.unregister();

		})
			.untilNoMoreServiceEventWithin(100l)// main stop-condition criteria
			.assertWithTimeoutThat(3000)// timeout to granite a stop
			.hasNoThrowable()// not exception thrown while executed
			.isNotTimedOut()// check that NOT timed-out
			.hasServiceEventsInOrder(//
				listOf(ServiceEventConditions.matches(ServiceEvent.REGISTERED, "(k=v0)"), //
					ServiceEventConditions.typeModifiedAndObjectClass(Serializable.class), //
					ServiceEventConditions.typeUnregisteringAndObjectClass(Serializable.class)))//
			.hasServiceEventsInExactOrder(//
				listOf(ServiceEventConditions.typeRegisteredWith(Serializable.class, mapOf("k", "v0")), //
					ServiceEventConditions.typeModifiedAndObjectClass(Serializable.class), //
					ServiceEventConditions.typeModifiedAndObjectClass(Serializable.class), //
					ServiceEventConditions.matches(ServiceEvent.UNREGISTERING, Serializable.class, mapOf("k", "v2"))))//
			.hasAtLeastNServiceEventModifiedWith(2, Serializable.class)//
			.hasAtMostNServiceEventModifiedWith(2, Serializable.class)//
			.hasExactlyNServiceEventModifiedWith(2, Serializable.class)//
			.hasExactlyOneServiceEventRegisteredWith(Serializable.class)//
			.hasAtLeastOneServiceEventModifiedWith(Serializable.class)//
			.hasAtLeastOneServiceEventUnregisteringWith(Serializable.class)//
			.hasAtLeastOneServiceEventWith(ServiceEvent.REGISTERED, "(k=v0)")//
			.hasAtLeastOneServiceEventRegisteredWith(Serializable.class, mapOf("k", "v0"))//
			.hasAtLeastOneServiceEventModifiedWith(Serializable.class, mapOf("k", "v1"))//
			.hasAtLeastOneServiceEventUnregisteringWith(Serializable.class, mapOf("k", "v2"))//
			.hasAtLeastOneServiceEventWith(ServiceEvent.REGISTERED, Serializable.class)

			.hasNoServiceEventWith(ServiceEvent.REGISTERED, "(k=v4)");

```


bjhargrave:
Can you rebase on main and also update the copyright comments to match the new form for this Eclipse project?

stbischof:
@bjhargrave rebase on main and also switched the copyright header to Eclipse.

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/event/EventRecordingAssert.java
```diff
@@ -0,0 +1,106 @@
...
+		int timeout) throws Exception {
+		return assertThat(execute, hasTypeAnd(FrameworkEvent.class, predicate), timeout);
+	}
+
+	/**
+	 * @param execute - the action that will be invoked.
+	 * @param predicate - Predicate that would be tested against the Events
+	 * @param timeout - the time in that the matches must happened
+	 */
+	public static EventRecordingAssert assertThat(Runnable execute, Predicate<Object> predicate, int timeout)
```
bjhargrave, line 83: The predicate argument should be typed `Predicate<?>` if you don't care what the type for the predicate input is. Using `Predicate<Object>` means that people cannot easily using a method reference to an existing method whose argument is not Object.

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/event/EventRecordingAssert.java
```diff
@@ -0,0 +1,106 @@
...
+	 */
+	public static EventRecordingAssert assertThat(Runnable execute, int timeout) throws Exception {
+		return assertThat(execute, (se) -> false, timeout);
+	}
+
+	/**
+	 * @param execute - the action that will be invoked.
+	 */
+	public static EventRecordingAssert assertThat(Runnable execute) throws Exception {
+		return assertThat(execute, 200);
```
bjhargrave, line 104: Where does `200` come from? It seems wrong to hard code a timeout value in an API. Perhaps we remove this method?

org.osgi.test.common/src/main/java/org/osgi/test/common/event/CloneUtil.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.Map;
+import java.util.stream.Stream;
+
+import org.osgi.framework.Bundle;
+import org.osgi.framework.BundleEvent;
+import org.osgi.framework.FrameworkEvent;
+import org.osgi.framework.ServiceEvent;
+import org.osgi.framework.ServiceReference;
+
+class CloneUtil {
```
bjhargrave, line 34: Since this is a package-private class, we should probably make the methods all package-private also.

bjhargrave, line 34: Why are we cloning things here?

org.osgi.test.common/src/main/java/org/osgi/test/common/event/CloneUtil.java
```diff
@@ -0,0 +1,114 @@
...
+					}
+					if (method.getName()
+						.equals("getBundle")) {
+						return bundle;
+					}
+					if (method.getName()
+						.equals("getUsingBundles")) {
+						return usingBundles;
+					}
+					return method.invoke(proxy, args);
```
bjhargrave, line 84: Isn't this an infinite recursion? You call the same method again on the same proxy object?

org.osgi.test.common/src/main/java/org/osgi/test/common/event/CloneUtil.java
```diff
@@ -0,0 +1,114 @@
...
+			public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+				if (method.getName()
+					.equals("getState")) {
+					return state;
+				}
+				if (method.getName()
+					.equals("getLastModified")) {
+					return lastModified;
+				}
+				return method.invoke(proxy, args);
```
bjhargrave, line 109: Infinite recursion?

org.osgi.test.common/src/main/java/org/osgi/test/common/event/EventRecording.java
```diff
@@ -0,0 +1,193 @@
...
+public interface EventRecording {
+	boolean isTimedOut();
+
+	List<TimedEvent<?>> timedEvents();
+
+	@SuppressWarnings("unchecked")
+	default <T> List<TimedEvent<T>> timedEvents(Class<T> clazz) {
+		return timedEvents().stream()
+			.filter((te) -> clazz.isInstance(te.getEvent()))
+			.map((te) -> (TimedEvent<T>) te)
```
bjhargrave, line 48: This method is a no-op. You want to assign the filtered collection to a variable of `List<TimedEvent<T>>` and make the case on that assignment.

```java
List<TimedEvent<T>> result = (List<TimedEvent<T>>) timedEvents().stream()
	.filter((te) -> clazz.isInstance(te.getEvent()))
	.collect(Collectors.toList());
return result;
```

org.osgi.test.common/src/main/java/org/osgi/test/common/event/EventRecording.java
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	default List<?> events() {
+		return timedEvents().stream()
+			.map(TimedEvent::getEvent)
+			.collect(Collectors.toList());
+	}
+
+	default <T> List<T> events(Class<T> clazz) {
+		return timedEvents(clazz).stream()
```
bjhargrave, line 59: Don't stream into a list and then restream into a new list.

org.osgi.test.common/src/main/java/org/osgi/test/common/event/EventRecording.java
```diff
@@ -0,0 +1,193 @@
...
+
+	default List<TimedEvent<FrameworkEvent>> timedFrameworkEvents() {
+		return timedEvents(FrameworkEvent.class);
+	}
+
+	static class EventsRecordingImpl implements EventRecording {
+		private List<TimedEvent<?>>	events;
+		private boolean				timedOut;
+
+		private EventsRecordingImpl() {}
```
bjhargrave, line 93: What is the purpose of this constructor?

org.osgi.test.common/src/main/java/org/osgi/test/common/event/EventRecording.java
```diff
@@ -0,0 +1,193 @@
...
+		}
+
+		@Override
+		public String toString() {
+			return "TimedEvent [event=" + event + ", instant=" + instant + "]";
+		}
+
+	}
+
+	static EventRecording record(BundleContext bc, Runnable action, Predicate<Object> predicate, int timeout)
```
bjhargrave, line 142: Use `Predicate<?>`

org.osgi.test.common/src/main/java/org/osgi/test/common/event/EventRecording.java
```diff
@@ -0,0 +1,193 @@
...
+		bc.addFrameworkListener(fListener);
+		bc.addBundleListener(bListener);
+		bc.addServiceListener(sListener);
+		action.run();
+		boolean timedOut = !latch.await(timeout, TimeUnit.MILLISECONDS);
+		bc.removeServiceListener(sListener);
+		bc.removeBundleListener(bListener);
+		bc.removeFrameworkListener(fListener);
+
+		List<TimedEvent<?>> objects = Collections.unmodifiableList(events);
```
bjhargrave, line 189: You probably want to wrap the original list rather than the synchronized wrapping.

bjhargrave (commented):
Initial round of comments.

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/event/AbstractEventRecordingAssert.java
```diff
@@ -0,0 +1,135 @@
...
+			.get(firstElementIndex);
+
+		TimedEvent<?> timedSecond = actual.timedEvents()
+			.get(secondElementIndex);
+
+		return createDurationAssert(timedFirst, timedSecond);
+	}
+
+	private DurationAssert createDurationAssert(TimedEvent<?> timedFirst, TimedEvent<?> timedSecond) {
+		return new DurationAssert(Duration.between(timedFirst.getInstant(), timedSecond.getInstant()));
```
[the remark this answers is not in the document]
stbischof: // TODO: use assertThat() when assertj-core 3.20.0 is relased
https://github.com/assertj/assertj-core/blame/main/src/main/java/org/assertj/core/api/Assertions.java#L972

org.osgi.test.common/src/main/java/org/osgi/test/common/event/CloneUtil.java
stbischof: done

stbischof: Because we want to freeze the
Bundle: state and lastModified
ServiceReference: Properties and Bundle
from the moment of the event and check this later.
If we do not copy this,it could change and a later the test would give the wrong result.

org.osgi.test.common/src/main/java/org/osgi/test/common/event/CloneUtil.java
stbischof: SURE! It must use the origin ServiceReference.

org.osgi.test.common/src/main/java/org/osgi/test/common/event/CloneUtil.java
stbischof: SURE! It must use the origin Bundle.

org.osgi.test.common/src/main/java/org/osgi/test/common/event/EventRecording.java
stbischof: this would result in a compile error
```
Cannot cast from List<EventRecording.TimedEvent<?>> to List<EventRecording.TimedEvent<T>>
```

bjhargrave: So you may need to cast to a raw List first. And also suppress the raw warning.

```java
List<TimedEvent<T>> result = (List<TimedEvent<T>>)(List) timedEvents().stream()
	.filter((te) -> clazz.isInstance(te.getEvent()))
	.collect(Collectors.toList());
return result;
```

org.osgi.test.common/src/main/java/org/osgi/test/common/event/EventRecording.java
stbischof: removed

org.osgi.test.common/src/main/java/org/osgi/test/common/event/EventRecording.java
stbischof: thx. changed

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/event/AbstractEventRecordingAssert.java
```diff
@@ -0,0 +1,135 @@
...
+		if (actual.isTimedOut()) {
+			throw failure("%nExpecting%n  <%s>%nto not be a timedOut, but it was", actual);
+		}
+		return myself;
+	}
+
+	public <T> ListAssert<T> hasEventsThat(Class<T> clazz) {
+		isNotNull();
+		List<T> list = actual.events(clazz);
+		return new ListAssert<T>(list);
```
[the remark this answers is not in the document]
stbischof: done

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/event/AbstractEventRecordingAssert.java
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	public <T> ListAssert<TimedEvent<T>> hasTimedEventsThat(Class<T> clazz) {
+		isNotNull();
+		List<TimedEvent<T>> list = actual.timedEvents(clazz);
+		return new ListAssert<TimedEvent<T>>(list);
```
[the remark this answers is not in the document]
stbischof: done

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/event/EventRecordingAssert.java
stbischof: I used `Predicate<? super Object>` because `Predicate<?>` will result in a compilation error on EventRecording.record.
```
The method test(capture#13-of ?) in the type Predicate<capture#13-of ?> is not applicable for the arguments (BundleEvent)	EventRecording.java
```

is this OK for you?

org.osgi.test.common/src/main/java/org/osgi/test/common/event/EventRecording.java
stbischof: added an additional `default <T> Stream<TimedEvent<T>> timedEventsStream(Class<T> clazz)`

org.osgi.test.common/src/main/java/org/osgi/test/common/event/EventRecording.java
stbischof: I used Predicate<? super Object>

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/event/EventRecordingAssert.java
bjhargrave: I don't think `Predicate<? super Object>` is helpful. I can't pass a `Predicate<BundleEvent>` argument. YOu may need some casting in the method?

org.osgi.test.common/src/main/java/org/osgi/test/common/event/EventRecording.java
stbischof: double cast.. first time in my life!

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/event/EventRecordingAssert.java
stbischof: thats my last idea;
https://github.com/osgi/osgi-test/pull/257/commits/36aea8fff80fc53063f0e3180940b37cba7d139f

stbischof:
@bjhargrave I fixes you initial comments.

stbischof:
OK, your question about the CloneUtil took a bit longer:
ServiceEvents abd BundleEvents must be immutable.
https://github.com/osgi/osgi/pull/265

stbischof:
I worked on this the lost weeks - now we have a much more fluent API.

pom.xml
```diff
@@ -47,6 +47,7 @@
 	<modules>
 		<module>org.osgi.test.common</module>
 		<module>org.osgi.test.assertj.framework</module>
+		<module>org.osgi.test.assertj.monitoring</module>
```
bjhargrave, line 50: Also need to add to BOM.

stbischof: added

stbischof:
I worked on this the lost weeks - now we have a much more fluent API.

Main Interfaces are:
* MonitoringAssertion.java
* MonitoringAssertionResult.java

The content of package `NotPartOfPR` will be removed and is handled in other PR's #256 #258

org.osgi.test.assertj.monitoring/src/main/java/org/osgi/test/assertj/monitoring/internal/MonitoringAssertionImpl.java
```diff
@@ -0,0 +1,297 @@
...
+				return runner.await();
+			} catch (InterruptedException e) {
+				e.printStackTrace();
+				return false;
+			}
+
+		}
+	}
+
+	private class Runner {
```
kriegfrj, line 274: I don't understand why this class is called "Runner" - it doesn't actually run anything?

org.osgi.test.assertj.monitoring/src/test/java/org/osgi/test/assertj/monitoring/internal/EventRecordingTest.java
```diff
@@ -0,0 +1,134 @@
...
+			Object listener = invocation.getArgument(0);
+			serviceListeners.add((ServiceListener) listener);
+			return null;
+		}).when(bundleContext).addServiceListener(Mockito.any(ServiceListener.class));
+		doAnswer(invocation -> {
+			Object listener = invocation.getArgument(0);
+			serviceListeners.remove(listener);
+			return null;
+		}).when(bundleContext).removeServiceListener(Mockito.any(ServiceListener.class));
+
```
kriegfrj, line 78: When your mock setup gets this complicated, it is much easier, more readable, more type-safe and more maintainable to directly implement the `BundleContext` interface as a concrete class. Especially as you are not using Mockito's `verify()` method anywhere.

org.osgi.test.assertj.monitoring/src/main/java/org/osgi/test/assertj/monitoring/internal/MonitoringAssertionImpl.java
```diff
@@ -0,0 +1,297 @@
...
+			try {
+				return runner.await();
+			} catch (InterruptedException e) {
+				e.printStackTrace();
+				return false;
+			}
+
+		}
+	}
+
```
kriegfrj, line 273: I can't see how this class + the "Runner" class can possibly work as it has been written? The same predicate instance has to be called twice within the timeout window. How is the same predicate instance going to get called twice? And at the end of it, there is no check to see if there were any other events while it was waiting.

org.osgi.test.assertj.monitoring/src/main/java/org/osgi/test/assertj/monitoring/internal/MonitoringAssertionImpl.java
```diff
@@ -0,0 +1,297 @@
...
+	 * @param execute   - the action that will be invoked.
+	 * @param predicate - Predicate that would be tested against the Events
+	 * @param timeout   - the time in that the matches must happened
+	 */
+	private static MonitoringResultAssertionImpl call(ThrowingCallable execute, Predicate<?> predicate, int timeout) {
+		BundleContext bc = FrameworkUtil.getBundle(MonitoringResultAssertionImpl.class).getBundleContext();
+
+		EventRecording eventRecording = EventRecording.record(bc, execute, predicate, timeout);
+		return new MonitoringResultAssertionImpl(eventRecording, MonitoringResultAssertionImpl.class);
+	}
```
kriegfrj, line 58: It seems to me that the idea behind `MonitoringResultAssertion` is to monitor events in real time (hence the predicates that have timeouts and suggest waiting). However, `EventRecording.record()` will run the `ThrowingCallable` to completion before it returns, so the entire event sequence will have already played out before the assertion instance is even created.

org.osgi.test.assertj.monitoring/src/main/java/org/osgi/test/assertj/monitoring/internal/AbstractMonitoringAssertion.java
```diff
@@ -0,0 +1,787 @@
...
+import org.osgi.test.assertj.NotPartOfPR.Conditions;
+import org.osgi.test.assertj.bundleevent.BundleEventAssert;
+import org.osgi.test.assertj.frameworkevent.FrameworkEventAssert;
+import org.osgi.test.assertj.monitoring.MonitoringAssertionResult;
+import org.osgi.test.assertj.monitoring.TimedEvent;
+import org.osgi.test.assertj.serviceevent.ServiceEventAssert;
+import org.osgi.test.common.bitmaps.ServiceEventType;
+import org.osgi.test.common.dictionary.Dictionaries;
+
+abstract class AbstractMonitoringAssertion<SELF extends AbstractMonitoringAssertion<SELF, ACTUAL>, ACTUAL extends EventRecording>
```
kriegfrj, line 52: For consistency, should be called `AbstractEventRecordingAssert`.

org.osgi.test.assertj.monitoring/src/main/java/org/osgi/test/assertj/monitoring/MonitoringAssertionResult.java
```diff
@@ -0,0 +1,661 @@
...
+import org.osgi.framework.InvalidSyntaxException;
+import org.osgi.framework.ServiceEvent;
+import org.osgi.test.assertj.bundleevent.BundleEventAssert;
+import org.osgi.test.assertj.frameworkevent.FrameworkEventAssert;
+import org.osgi.test.assertj.serviceevent.ServiceEventAssert;
+
+/**
+ * The Interface RuntimeMonitoringResultAssert.
+ */
+public interface MonitoringAssertionResult {
```
kriegfrj, line 40: What is the thinking behind this interface?

kriegfrj (changes requested):
I've had a bit of a pass at this. I think there is significant work to do to make it useful. I'll see how you respond to the first pass and if there's any significant change in direction I'll come back for more. :smile:

org.osgi.test.assertj.monitoring/src/main/java/org/osgi/test/assertj/monitoring/MonitoringAssertionResult.java
stbischof: Gives access to assertion methods. All assertions will be executed on the recorded events.

org.osgi.test.assertj.monitoring/src/main/java/org/osgi/test/assertj/monitoring/internal/MonitoringAssertionImpl.java
stbischof: This is what it should do.

Run the `ThrowingCallable`  complete and record everything happened in the `ThrowingCallable` but stop directly after the call if the predicate succeeds (or wait until the timeout)  that the predicate succeeds maybe later.

this way it is possible to do a kind of inactivity check where we do some more complex predicates
```
	/**
	 * Until no more service event within given time is fired.
	 *
	 * @param millis      the time in milliseconds
	 * @param objectClass the objectClass
	 * @return MonitoringAssertTimeoutStep - to define the
	 *         observation-timeout.
	 */
	MonitoringAssertionTimeoutStep untilNoMoreServiceEventOfClassWithin(long millis, Class<?> objectClass);
```

stbischof:
I will close this and come back if a have a simpler solution.

PR #288, opened 2021-03-04
Bill of Materials

stbischof:
Let other projects easily import all OSGi-ttest dependencies as Bill of Materials.

### DIFF
--- a/org.osgi.test.bom/.project
+++ b/org.osgi.test.bom/.project
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.osgi.test.bom</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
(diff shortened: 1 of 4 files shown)


org.osgi.test.bom/.project
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>org.osgi.test.common</name>
```
rotty3000, line 3: This needs to be corrected right?

rotty3000 (changes requested):
Otherwise LGTM

org.osgi.test.bom/.project
stbischof: org.osgi.test.bom   right!

rotty3000: nice!

rotty3000 (approved):

rotty3000:
Thanks @stbischof, this is a nice addition.

bjhargrave:
Well, the copyright notice is wrong. I just updated all the notices for the project for Eclipse. So we need this one to match.

PR #291, opened 2021-03-05
bom is broken

### LINKED ISSUE #290: bom is broken
The bom project just added seems rather broken. 

When I build from the command line

```
[INFO] ---------------------< org.osgi:org.osgi.test.bom >---------------------
[INFO] Building OSGi-Test Bill of Materials ${revision}                   [8/8]
[INFO] --------------------------------[ pom ]---------------------------------
[INFO] 
[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ org.osgi.test.bom ---
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] org.osgi:org.osgi.test.parent 0.11.0-SNAPSHOT ...... SUCCESS [  1.627 s]
[INFO] org.osgi:org.osgi.test.common 0.11.0-SNAPSHOT ...... SUCCESS [ 11.846 s]
[INFO] org.osgi:org.osgi.test.assertj.framework 0.11.0-SNAPSHOT SUCCESS [  7.091 s]
[INFO] org.osgi:org.osgi.test.assertj.promise 0.11.0-SNAPSHOT SUCCESS [  7.939 s]
[INFO] org.osgi:org.osgi.test.junit4 0.11.0-SNAPSHOT ...... SUCCESS [  4.440 s]
[INFO] org.osgi:org.osgi.test.junit5 0.11.0-SNAPSHOT ...... SUCCESS [ 12.422 s]
[INFO] org.osgi:org.osgi.test.junit5.cm 0.11.0-SNAPSHOT ... SUCCESS [ 13.901 s]
[INFO] OSGi-Test Bill of Materials ${revision} ............ SUCCESS [  0.001 s]
```

The `${revision}` in the output seems like something bad. The bom does not reference the parent project and so does not know what any of the properties like `revision` are. Also, don't we need some tests?

It seems like this was merged prematurely.

rotty3000:
fixes #290

### DIFF
--- a/org.osgi.test.bom/pom.xml
+++ b/org.osgi.test.bom/pom.xml
@@ -1,115 +1,87 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-	* Copyright (c) OSGi Alliance (2021). All Rights Reserved.
-	*
-	* Licensed under the Apache License, Version 2.0 (the "License");
-	* you may not use this file except in compliance with the License.
-	* You may obtain a copy of the License at
-	*
-	* http://www.apache.org/licenses/LICENSE-2.0
-	*
-	* Unless required by applicable law or agreed to in writing, software
-	* distributed under the License is distributed on an "AS IS" BASIS,
-	* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-	* See the License for the specific language governing permissions and
-	* limitations under the License.
+	Copyright (c) Contributors to the Eclipse Foundation
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+		http://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+
+	SPDX-License-Identifier: Apache-2.0
 -->
 <project xmlns="http://maven.apache.org/POM/4.0.0"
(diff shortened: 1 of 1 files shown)


org.osgi.test.bom/pom.xml
```diff
@@ -162,4 +134,16 @@
...
 		</dependencies>
 	</dependencyManagement>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.codehaus.mojo</groupId>
+				<artifactId>flatten-maven-plugin</artifactId>
+				<configuration>
+					<flattenMode>bom</flattenMode>
```
bjhargrave, line 144: Yes!

org.osgi.test.bom/pom.xml
```diff
@@ -1,115 +1,87 @@
...
+	<scm>
+		<url>https://github.com/osgi/osgi-test</url>
+		<connection>scm:git:https://github.com/osgi/osgi-test.git</connection>
+		<developerConnection>scm:git:git@github.com:osgi/osgi-test.git</developerConnection>
+	</scm>
+
+	<packaging>pom</packaging>
+
+	<properties>
+		<bnd.resolve.skip>true</bnd.resolve.skip>
```
bjhargrave, line 44: I am concerned that while this is necessary to build in our build, this will bleed through into all users of the bom? Or is that an incorrect assumption?

rotty3000: I'll check

rotty3000: In my test it did not appear to have any side effect. And I don't believe properties of the Bom are ever consumed downstream from a bom.

rotty3000: However if there is real concern we can directly configure the plugin instead of using the property.

bjhargrave:
We need to merge this since the main build is broken: https://github.com/osgi/osgi-test/runs/2056604113?check_suite_focus=true#step:7:1307

PR #305, opened 2021-04-21
Avoid caching a potentially stale BundleContext

### LINKED ISSUE #212: @InjectBundleContext lifecycle problems
## Background
The "normal" way to use the `bnd-testing-maven-plugin` is to package up all the unit tests and OSGi tests in a single test jar which is a fragment of the bundle being tested. This allows the unit tests to be in the same package(s) as the classes under test and therefore default visibility methods can be tested. The consequence of this is that in the OSGi tests the `BundleContext` for the tester bundle is the same as for the bundle being tested.

## The Problem
In some of my OSGi tests I need to ensure a "clean" start of the bundle under test to avoid residual state (e.g. historical monitoring data). The simplest way to do this is to restart the bundle under test. This invalidates the `BundleContext` injected into my test and creates a new one. To ensure that I don't use a stale `BundleContext` I inject a new one into my test class using an `@BeforeEach` method, unfortunately the object my test is injected with appears to be the old invalid `BundleContext` and throws Exceptions when I try to use it in the next test.

## Proposed solution
Each time the `BundleContext` is injected into a lifecycle it should be obtained from the tester bundle, and not cached and reused.

timothyjward:
Fixes #212

Signed-off-by: Tim Ward <timothyjward@apache.org>

### DIFF
--- a/org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/context/BundleContextRuleTest.java
+++ b/org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/context/BundleContextRuleTest.java
@@ -19,6 +19,7 @@
 package org.osgi.test.junit4.context;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.fail;
 import static org.osgi.framework.Bundle.INSTALLED;
 import static org.osgi.framework.Bundle.UNINSTALLED;
 import static org.osgi.test.junit4.TestUtil.getBundle;
@@ -32,6 +33,7 @@ import org.junit.rules.TestName;
 import org.osgi.framework.Bundle;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.BundleEvent;
+import org.osgi.framework.BundleException;
 import org.osgi.framework.BundleListener;
 import org.osgi.framework.FrameworkUtil;
 import org.osgi.framework.ServiceObjects;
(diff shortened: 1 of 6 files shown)


org.osgi.test.junit4/src/test/java/org/osgi/test/junit4/context/BundleContextRuleTest.java
```diff
@@ -230,14 +232,37 @@ public void cleansUpGottenServiceObjects() throws Exception {
...
+		try {
+			// Should throw an exception
+			closeableBC.getServiceReference(Object.class);
+			fail("Expected an IllegalStateException");
+		} catch (IllegalStateException ise) {}
+
+		upstream.start();
+
+		// Should not throw an exception
+		closeableBC.getServiceReference(Object.class);
```
kriegfrj, line 266: What is going to happen here when you call `close()`? It seems to me the following:

* First call to `getServiceReference()` is on the original `BundleContext` BC1. CBC will cache the service reference obtained thereby so that it can unget it when close() is called..
* Restart will invalidate the original `BundleContext` BC1 (and the service reference obtained above) and create a new one BC2.
* Second call to `getServiceReference()` will be on BC2. CBC will cache this service reference too.
*  When `close()` is called, it will try and unget both of the services references on BC2.

This seems problematic to me. I think that, if we're going to go down this route (which I'm still not 100% conviced of) the CloseableBundleContext needs to invalidate all of its cached references when the Bundle stops..

kriegfrj (commented):
This is not a detailed review. It is mainly to highlight what I think is the root issue.

My understanding of the OSGi spec is that when a bundle is stopped, code that lives in that bundle should stop executing. The root cause of this problem with the stale BundleContext is that this rule is being violated - the bundle that contains the test code is being restarted because it is wired to the bundle-under-test, so when you stop the bundle-under-test, the test bundle stops too (which is when the BundleContext becomes invalid).

I understand the need to restart the bundle-under-test. In this case, I think it is better to put your test in a Bundle that is not hard wired to the bundle-under-test. This is doable if you only need to test the external effects of the bundle restart (eg, log files being emptied). Alternatively, your test bundle can use reflection to peek inside the bundle-under-test at the stuff you need to look at and avoid creating an OSGi wire link to the bundle-under-test, so that when the bundle-under-test restarts the test bundle does not (and hence the BundleContext does not become invalid).

That's how I feel at the moment, but I realise I'm not the most expert at OSGi of those here so I'm happy to defer to more seasoned opinions if that's where we go.

bjhargrave:
So the point of this is that `CloseableBundleContext` tracks not a `BundleContext` but a `Bundle` and returns the current `BundleContext` for the tracked `Bundle` whenever asked?

If that is so, then the `CloseableBundleContext` needs to watch the tracked bundle so that it can "flush" any tracked BundleContext-related information (service registrations, listeners, services, service objects) when the bundle is stopped since the framework will have done that for the stopped bundle. The only thing tracked by `CloseableBundleContext` that can "survive" the stopping of the tracked bundle is the set of installed bundles.

So this change would also need to properly handle stopping of the tracked bundle otherwise information will "leak" across different startings of the tracked bundle.

kriegfrj:
Closed per discussion in #212.

Issue #148, opened 2020-06-26
Service testing sometimes need to test what happens when a service is released


bjhargrave:
In trying this project in an OSGi compliance test, I came across a test case which uses a service to get an object and then needs to release the service to confirm some behavior is performed on the object (closing the object in this case).

But the ServiceExtension support does not provide anyway to release a service. You can inject it as a ServiceAware, but ServiceAware does not provide access to the underlying close method on ServiceConfiguration (which implements ServiceAware).

Perhaps ServiceAware can add a close method to enable tests to release the service early to confirm certain behavior?

PR #323, opened 2021-06-03
Some fixes to get ready for a release

bjhargrave:
This PR updates the MapStream code and also fixes https://github.com/osgi/osgi-test/issues/148.


org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/ServiceExtension_ServiceAwareTest.java
```diff
@@ -310,4 +310,56 @@ public void matchMultiple() throws Exception {
...
+				.as("size %s", serviceAware)
+				.isEqualTo(1);
+			softly.assertThatCode(() -> serviceAware.close())
+				.doesNotThrowAnyException();
+			softly.assertThat(serviceAware.isEmpty())
+				.as("isEmpty %s", serviceAware)
+				.isTrue();
+			softly.assertThat(serviceAware.size())
+				.as("size %s", serviceAware)
+				.isEqualTo(0);
```
kriegfrj, line 344: When we see code like this it's often a hint that we should have an AssertJ class for the type under test. :smile:

I have pondered adding an AssertJ class for `ServiceAware`; if it would have utility beyond osgi-test's own internal test cases (I suspect it might) we should look at it.

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/AbstractServiceExtensionTest.java
```diff
@@ -121,7 +121,7 @@ public void afterEach() throws Exception {
 		checkClass(testClass);
 		try {
 			return executor.schedule(() -> assertThatTest(testClass), delay, TimeUnit.MILLISECONDS)
-				.get(delay + 200000, TimeUnit.MILLISECONDS);
+				.get(200000L + delay, TimeUnit.MILLISECONDS);
```
kriegfrj, line 124: I don't object to this change but for my own personal education purposes I am curious as to the reasoning.

org.osgi.test.junit5/test.bndrun
```diff
@@ -44,17 +44,17 @@
...
+	junit-jupiter-engine;version='[5.7.2,5.7.3)',\
+	junit-jupiter-params;version='[5.7.2,5.7.3)',\
+	junit-platform-commons;version='[1.7.2,1.7.3)',\
+	junit-platform-engine;version='[1.7.2,1.7.3)',\
+	junit-platform-launcher;version='[1.7.2,1.7.3)',\
+	junit-platform-testkit;version='[1.7.2,1.7.3)',\
+	net.bytebuddy.byte-buddy;version='[1.11.0,1.11.1)',\
+	net.bytebuddy.byte-buddy-agent;version='[1.11.0,1.11.1)',\
+	org.mockito.mockito-core;version='[3.10.0,3.10.1)',\
+	org.objenesis;version='[3.2.0,3.2.1)',\
```
kriegfrj, line 57: I recall that our strategy was generally to keep these on lower versions to verify that our code doesn't always require the latest and greatest; however we were happy to bump the version if there was an established need. I was wondering if there  was a particular need that you had established or else this was just a general catch-up?

org.osgi.test.common/src/main/java/org/osgi/test/common/service/ServiceAware.java
```diff
@@ -56,4 +56,5 @@
 
 	long getTimeout();
 
+	void close();
```
kriegfrj, line 59: 👍

kriegfrj (commented):
Nothing really major...

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/ServiceExtension_ServiceAwareTest.java
bjhargrave: Sounds like an interesting addition for version 1.1 (after this release!).

org.osgi.test.junit5/src/test/java/org/osgi/test/junit5/service/AbstractServiceExtensionTest.java
bjhargrave: The argument to `get` must be a long. The code had an int+int which makes an int which then must be widened to a long. If delay was, for example, Integer.MAX_VALUE then the int+int would overflow an int. By making the expression long+int, the compiler will first widen the int delay argument to long and then add to the constant long. So we never can overflow. I just happened to see this when looking at the code. I always try to use `L` for long constants and `D` for double constants to be explicit and avoid making the compiler infer what I mean.

org.osgi.test.junit5/test.bndrun
bjhargrave: I have been merging the dependabot PRs which don't handle these file. This is just catching up with the values in the pom.

PR #326, opened 2021-06-06
[AssertJ] add ServiceRegistrationAssert

stbischof:
* add ServiceRegistrationAssert
* minor fixes

### DIFF
--- a/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
+++ b/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/servicereference/AbstractServiceReferenceAssert.java
@@ -44,7 +44,8 @@ public abstract class AbstractServiceReferenceAssert<SELF extends AbstractServic
 	}
 
 	public DictionaryAssert<String, Object> hasServicePropertiesThat() {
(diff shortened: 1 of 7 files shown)


stbischof:
Thank you @kriegfrj  for the additional Test.

kriegfrj:
> Thank you @kriegfrj for the additional Test.

No problem! The purpose of the test is to ensure that the actual value passed to the child assert is the same value that comes from the parent's property.

I'm not quite sure how I managed to push it directly to your branch, though? I meant to push it to my own fork and do a PR for you to look at so you could add it yourself, but I pushed it to your fork by accident. I wouldn't have thought I'd have permission to push to your repo. Is that some GitHub feature that I can push to it because it was part of a pull request to a repo that I have commit permissions on?

stbischof:
```
If checked, users with write access to osgi/osgi-test can add new commits to your ServiceRegAssert branch. You can always change this setting later.

Maintainers could potentially edit this repository's workflows to reveal values of secrets and gain access to other branches.
```

bjhargrave:
See <https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork>. I take advantage of this from time-to-time when it is easier to make a change to a PR myself than to explain what changes are needed.

PR #330, opened 2021-06-07
Bitmap - typeMatchedMask  methods

stbischof:


org.osgi.test.common/src/main/java/org/osgi/test/common/bitmaps/Bitmap.java
```diff
@@ -126,4 +127,12 @@ public String maskToString(int mask) {
...
 		return mask + ":" + bits.collect(Collectors.joining(" | "));
 	}
+
+	public static BiPredicate<Integer, Integer> typeMatchesMask() {
+		return (type, mask) -> (type & mask) != 0;
+	}
+
+	public static boolean typeMatchesMask(int type, int mask) {
+		return typeMatchesMask().test(type, mask);
+	}
```
kriegfrj, line 137: I don't think that `typeMatchesMask()` is necessary. I would simply have:

```java
public static boolean typeMatchesMask(int type, int mask) {
		return type & mask != 0;
}
```

Then if you need a `BiPredicate` somewhere, you can simply use `Bitmap::typeMatchesMask`.

Haven't said that, the function is pretty trivial so I'm not sure that we're gaining a lot over simply inlining the boolean operation? Do you have a specific use case in mind so I can get a feel for how useful it is?

kriegfrj (commented):
Not 100% convinced...

org.osgi.test.common/src/main/java/org/osgi/test/common/bitmaps/Bitmap.java
stbischof: For me its not as trivial ;-) So it was easier to extract it and just use am readable Function

bjhargrave: If you really want to do this in the form of a predicate, I would author a BiIntPredicate (see IntPredicate in Java) to avoid unnecessary boxing and unboxing. 



> For me its not as trivial to read. ;-) So it was easier to extract it and just use am readable Function.

I don't see any use of the predicate in that link.

stbischof: i removed the bipredicat. since some changes i did in assertj the bipredicate is not longer necessary.

stbischof:
to just add this one static method would be enough for me in this PR.

org.osgi.test.common/src/main/java/org/osgi/test/common/bitmaps/Bitmap.java
```diff
@@ -126,4 +126,8 @@ public String maskToString(int mask) {
 		}
 		return mask + ":" + bits.collect(Collectors.joining(" | "));
 	}
+
+	public static boolean typeMatchesMask(int type, int mask) {
```
bjhargrave, line 130: Shouldn't you bump the package version to 1.1?

stbischof: sure. i pushed to the wrong repo.

bjhargrave (approved):

PR #331, opened 2021-06-07
[cm] Updated runbundles

kriegfrj:
Latest resolve doesn't include org.osgi.service.cm as org.apache.felix.configadmin includes those packages.

### DIFF
--- a/org.osgi.test.junit5.cm/test.bndrun
+++ b/org.osgi.test.junit5.cm/test.bndrun
@@ -51,7 +51,6 @@
 	junit-platform-launcher;version='[1.7.2,1.7.3)',\
 	org.apache.felix.configadmin;version='[1.9.22,1.9.23)',\
 	org.opentest4j;version='[1.2.0,1.2.1)',\
-	org.osgi.service.cm;version='[1.6.0,1.6.1)',\
 	org.osgi.test.assertj.framework;version='[1.1.0,1.1.1)',\
 	org.osgi.test.common;version='[1.1.0,1.1.1)',\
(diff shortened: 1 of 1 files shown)


bjhargrave (commented):
When I run `./mvnw clean  verify` in my system, it puts this line back. So I don't see that it should have been removed.

kriegfrj:
Hmmm, when I ran mvn verify it removed it. I'll double check. Certainly on manual inspection it doesn't appear to be needed for the reasons stated in the OP.

kriegfrj:
Ok, I have double-checked on a different machine (still Windows 10), in case that was the issue, and I have also tried on both Java 11 and 1.8. In every scenario, the resolver wants to remove `org.osgi.service.cm`.

My understanding upon manual inspection is that this is as it should be - the resolver should not be including `org.osgi.service.cm` because `org.apache.felix.configadmin` contains all of those dependencies. So I think the mystery is why is it being included on your system?

rotty3000:
We've been through similar debates before. The resolver results are not
idempotent. They are merely 'correct'. Any number of slight environmental
conditions might result in a variation of a decision which could influence
the outcome. The only way to assure the output is to more strictly control
the inputs. I.e. if the cm API bundle isn't needed, don't let it be part of
the resolution result by marking it as a 'provided' dependency and then
make sure the resolver configuration does not use the provided scope.

bjhargrave:
On my system, it is always added :-) 

I would rather err on the side of leaving it in. The most recent consensus is that impls should not export the spec packages they implements (although this is not in 100% agreement) but instead import the spec package.

PR #341, opened 2021-06-16
Assertions fixes for assertj 3.20

bjhargrave:


org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/bundle/AbstractBundleAssert.java
```diff
@@ -92,8 +92,8 @@ public SELF hasSymbolicName(String expected) {
 		return myself;
 	}
 
-	public AbstractStringAssert<?> hasSymbolicNameThat() {
-		return isNotNull().extracting(Bundle::getSymbolicName, STRING)
+	public StringAssert hasSymbolicNameThat() {
+		return (StringAssert) isNotNull().extracting(Bundle::getSymbolicName, STRING)
```
kriegfrj, line 96: What is the motivation for this change (and others like it)? It's not clear to me on the face of it that a change which requires introducing a cast is actually an improvement.

kriegfrj: Sorry, just saw 7037b18c7193f59c8cd8750251bf0bedeeeb90d0

bjhargrave: Yes, it was temporary to provide a nice commit to cherry-pick to release 1.0. I wanted to fix the API in our 1.0 release so we did not change API afterwards.

PR #357, opened 2021-07-15
Add the dynamic ServiceAware functionality to injected Lists

stbischof:
from `osgi-test` mailing list

```
I was expecting that if I would have an injected List<S> with S a certain type of service, and I would register a service using the BundleContext in the test, the List<S> would reflect the newly registered service. A ServiceAware<S> also does not seem to reflect the new situation.

Are my expectations wrong?
```

This PR could solve this Problem but will break an other existing test.

@bjhargrave what do you think about?

### DIFF
--- a/org.osgi.test.common/src/main/java/org/osgi/test/common/list/ListSupplierDelegate.java
+++ b/org.osgi.test.common/src/main/java/org/osgi/test/common/list/ListSupplierDelegate.java
@@ -1,0 +1,215 @@
+package org.osgi.test.common.list;
+
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.Objects;
+import java.util.Spliterator;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+import java.util.function.UnaryOperator;
+import java.util.stream.Stream;
+
+public class ListSupplierDelegate<E> implements List<E> {
+
+	private final Supplier<? extends List<E>> supplier;
+
+	public ListSupplierDelegate(Supplier<? extends List<E>> supplier) {
+		this.supplier = Objects.requireNonNull(supplier);
+	}
+
+	@Override
+	public boolean add(E e) {
+		return supplier.get()
+			.add(e);
+	}
+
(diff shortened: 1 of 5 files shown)


org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceAwareList.java
```diff
@@ -0,0 +1,215 @@
...
+import java.util.ListIterator;
+import java.util.Spliterator;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+import java.util.function.UnaryOperator;
+import java.util.stream.Stream;
+
+import org.osgi.test.common.service.ServiceAware;
+
+public class ServiceAwareList<E> implements List<E> {
```
bjhargrave, line 16: You could generalize this in to a `ListSupplierDelegate`.

```java
public class ListSupplierDelegate<E> implements List<E> {
    private final Supplier<? extends List<? extends E>> supplier;
    public ListSupplierDelegate(Supplier<? extends List<? extends E>> supplier) {
        this.supplier = Objects.requireNonNull(supplier);
    }
    @Override
	public void forEach(Consumer<? super E> action) {
		supplier.get()
			.forEach(action);
	}
```
 You create one with `new ListSupplierDelegate<>(configuration::getServices)`.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceAwareList.java
```diff
@@ -0,0 +1,215 @@
...
+
+	private ServiceAware<E> sa;
```
bjhargrave, line 18: This needs to be final

stbischof: ok

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceAwareList.java
stbischof: good idea

bjhargrave (approved):

bjhargrave:
Nice! Thanks.

PR #428, opened 2021-11-30
[assertj] Add global soft assertion entry point for framework

### LINKED ISSUE #115: [assertj] Global assert entry-points
Creating a new issue to capture discussion that began in a pull request here: https://github.com/osgi/osgi-test/pull/110#discussion_r424813029 Thought that the discussion was important enough that it be captured in a top-level issue for posterity rather than buried somewhere as an appendage to another PR.

In summary, we think it might be a good idea on balance to have an "uber soft assertion provider" interface which amalgamates all of the individual soft assertion providers into a single interface. Probably, the same arguments apply (to a lesser degree) to have an `Assertions` class that contains entry points for the hard assertions.

One question still to be resolved is where to put these uber-classes. Given the package versioning stability issues that @bjhargrave is justly concerned about, it would probably best if they were put in their own package. A couple of candidates:

* org.osgi.test.assertj
* org.osgi.test.assertj.api

_Edit: fixed link_

kriegfrj:
Closes #115.

### DIFF
--- a/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/FrameworkSoftAssertionsProvider.java
+++ b/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/FrameworkSoftAssertionsProvider.java
@@ -1,0 +1,20 @@
+package org.osgi.test.assertj;
+
+import org.osgi.test.assertj.bundle.BundleSoftAssertionsProvider;
+import org.osgi.test.assertj.bundlecontext.BundleContextSoftAssertionsProvider;
+import org.osgi.test.assertj.bundleevent.BundleEventSoftAssertionsProvider;
+import org.osgi.test.assertj.bundlereference.BundleReferenceSoftAssertionsProvider;
+import org.osgi.test.assertj.dictionary.DictionarySoftAssertionsProvider;
+import org.osgi.test.assertj.frameworkevent.FrameworkEventSoftAssertionsProvider;
+import org.osgi.test.assertj.serviceevent.ServiceEventSoftAssertionsProvider;
+import org.osgi.test.assertj.servicereference.ServiceReferenceSoftAssertionsProvider;
+import org.osgi.test.assertj.serviceregistration.ServiceRegistrationSoftAssertionsProvider;
+import org.osgi.test.assertj.version.VersionSoftAssertionsProvider;
+import org.osgi.test.assertj.versionrange.VersionRangeSoftAssertionsProvider;
+
+public interface FrameworkSoftAssertionsProvider
+	extends BundleSoftAssertionsProvider, BundleContextSoftAssertionsProvider, BundleEventSoftAssertionsProvider,
+	BundleReferenceSoftAssertionsProvider, DictionarySoftAssertionsProvider, FrameworkEventSoftAssertionsProvider,
+	ServiceEventSoftAssertionsProvider, ServiceReferenceSoftAssertionsProvider,
+	ServiceRegistrationSoftAssertionsProvider, VersionSoftAssertionsProvider, VersionRangeSoftAssertionsProvider {
+}
--- a/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/version/VersionSoftAssertionsProvider.java
+++ b/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/version/VersionSoftAssertionsProvider.java
@@ -16,6 +16,19 @@
  * SPDX-License-Identifier: Apache-2.0
  *******************************************************************************/
 
-@org.osgi.annotation.bundle.Export
-@org.osgi.annotation.versioning.Version("1.1.0")
 package org.osgi.test.assertj.version;
+
+import org.assertj.core.api.SoftAssertionsProvider;
+import org.osgi.framework.Version;
+
+public interface VersionSoftAssertionsProvider extends SoftAssertionsProvider {
+	/**
+	 * Create assertion for {@link org.osgi.framework.Version}.
+	 *
+	 * @param actual the actual value.
+	 * @return the created assertion object.
+	 */
+	default VersionAssert assertThat(Version actual) {
+		return proxy(VersionAssert.class, Version.class, actual);
+	}
+}
--- a/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/versionrange/VersionRangeSoftAssertionsProvider.java
+++ b/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/versionrange/VersionRangeSoftAssertionsProvider.java
@@ -16,6 +16,19 @@
  * SPDX-License-Identifier: Apache-2.0
  *******************************************************************************/
 
-@org.osgi.annotation.bundle.Export
-@org.osgi.annotation.versioning.Version("1.1.0")
-package org.osgi.test.assertj.version;
+package org.osgi.test.assertj.versionrange;
+
+import org.assertj.core.api.SoftAssertionsProvider;
+import org.osgi.framework.VersionRange;
+
+public interface VersionRangeSoftAssertionsProvider extends SoftAssertionsProvider {
+	/**
+	 * Create assertion for {@link org.osgi.framework.Version}.
+	 *
+	 * @param actual the actual value.
+	 * @return the created assertion object.
+	 */
+	default VersionRangeAssert assertThat(VersionRange actual) {
+		return proxy(VersionRangeAssert.class, VersionRange.class, actual);
+	}
+}
--- a/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/package-info.java
+++ b/org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/package-info.java
@@ -17,5 +17,5 @@
  *******************************************************************************/
 
 @org.osgi.annotation.bundle.Export
-@org.osgi.annotation.versioning.Version("1.1.0")
-package org.osgi.test.assertj.version;
+@org.osgi.annotation.versioning.Version("1.0.0")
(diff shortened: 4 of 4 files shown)


org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/FrameworkSoftAssertionsProvider.java
```diff
@@ -0,0 +1,20 @@
...
+import org.osgi.test.assertj.bundlereference.BundleReferenceSoftAssertionsProvider;
+import org.osgi.test.assertj.dictionary.DictionarySoftAssertionsProvider;
+import org.osgi.test.assertj.frameworkevent.FrameworkEventSoftAssertionsProvider;
+import org.osgi.test.assertj.serviceevent.ServiceEventSoftAssertionsProvider;
+import org.osgi.test.assertj.servicereference.ServiceReferenceSoftAssertionsProvider;
+import org.osgi.test.assertj.serviceregistration.ServiceRegistrationSoftAssertionsProvider;
+import org.osgi.test.assertj.version.VersionSoftAssertionsProvider;
+import org.osgi.test.assertj.versionrange.VersionRangeSoftAssertionsProvider;
+
+public interface FrameworkSoftAssertionsProvider
```
bjhargrave, line 15: I am not sure this is a great idea. From a semantic versioning perspective, each time a new provide is added, we need to bump the major version. From a cohesiveness perspective, these are completely unrelated.

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/version/VersionSoftAssertionsProvider.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *******************************************************************************/
+
+package org.osgi.test.assertj.version;
+
+import org.assertj.core.api.SoftAssertionsProvider;
+import org.osgi.framework.Version;
+
+public interface VersionSoftAssertionsProvider extends SoftAssertionsProvider {
```
bjhargrave, line 24: We should have some use of this in tests, no?

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/versionrange/VersionRangeSoftAssertionsProvider.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *******************************************************************************/
+
+package org.osgi.test.assertj.versionrange;
+
+import org.assertj.core.api.SoftAssertionsProvider;
+import org.osgi.framework.VersionRange;
+
+public interface VersionRangeSoftAssertionsProvider extends SoftAssertionsProvider {
```
bjhargrave, line 24: We should have some use of this in tests, no?

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/FrameworkSoftAssertionsProvider.java
kriegfrj: I thought we had discussed this [here](https://github.com/osgi/osgi-test/pull/110#discussion_r424813029)? To 
* Adding a new provider is backward compatible. As such, we should only need to bump the minor version. (I admit to not being a versioning expert so I prepare to stand corrected on this.)
* Now that we have `SoftAssertionsProvider`s that cover more-or-less the entirety of  `org.osgi.framework` package, the only time we're going to have add new ones is if/when new classes/interfaces are added to the `org.osgi.framework` package. You'd probably have a more direct appreciation for whether or not this is a frequent occurrence, but I can't see it happening that frequently.
* I don't agree with the statement that these are completely unrelated. The relation between them may be succinctly defined as "the set of all soft assertion providers for classes in the `org.osgi.framework` package". This set would seem to be no less cohesive than the `org.osgi.framework` package itself.

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/version/VersionSoftAssertionsProvider.java
kriegfrj: Yes.

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/versionrange/VersionRangeSoftAssertionsProvider.java
kriegfrj: Yes.

org.osgi.test.assertj.framework/src/main/java/org/osgi/test/assertj/FrameworkSoftAssertionsProvider.java
bjhargrave: We did discuss it in the mentioned issue and I said then I did not care for these uber-types. I am not sure what real utility they actually serve.

None of these interfaces are marked `ProviderType` which means they are `ConsumerType` and adding methods is a major change. If this is what we plan to do, we should properly mark the types.

If the types are not unrelated, then why are they all in different packages?

kriegfrj: > We did discuss it in the mentioned issue and I said then I did not care for these uber-types. I am not sure what real utility they actually serve.

Initially you said you didn't care for them but after our discussion you gave it a thumbs up [here](https://github.com/osgi/osgi-test/issues/115#issuecomment-631873635). If you've reverted, that's ok. I'll be honest, I haven't had a great need for this in my own testing as yet, so I'm happy to hold off on this PR.

> None of these interfaces are marked ProviderType which means they are ConsumerType and adding methods is a major change. If this is what we plan to do, we should properly mark the types.

So I stand corrected then. :smile: I'm going to have to process this information as I wasn't fully aware of this distinction or its ramifications. I feel that these interfaces are not really meant to be implemented by consumers of the AssertJ bundle and so we could justifiably mark them as `@ProviderType`. 

But I'm a bit surprised that adding a *default* method is considered a major change in any context - makes complete sense to me to make it major for an abstract method to be added, but the reasoning behind making it a major change for a default method is not so clear. (The only possible reason I can think of is that new method name might clash with an existing method name in your own implementation/subclass?)

> If the types are not unrelated, then why are they all in different packages?

They're all in the same sub-package though? There's degrees of relatedness. To say that I am more closely related to my siblings than my cousins is true, but that does not mean that I am unrelated to my cousins. In this case, the classes are cousins rather than siblings. This makes intuitive sense to me but perhaps this way of thinking doesn't transfer well to OSGi practice.

bjhargrave: Well they are all also in the `org` subpackage as is thousands (millions?) of other classes :-) Having the uber-soft assertion type in assertj-core works because there is one API package. We have many, even in the `assertj.framework` bundle.

`SoftAssertionsProvider` is an interface we don't control and is the super interface of all our soft assertions provider types. We cannot know what assertj will do with `SoftAssertionsProvider` in the future that could break our claim that our types are `@ProviderType`. (Of course we are always subject to changes made by assertj which have already been somewhat painful.)

I like the add of the missing individual soft assertion providers types here. But I am not sure we need to provide an uber-soft assertion provider type. People can easily make their own mixing in the things they care about.

kriegfrj: > Well they are all also in the org subpackage as is thousands (millions?) of other classes :-)

Well, I did say that there are *degrees* of relatedness. We're all related to each other via common ancestry too. You might be my 50th cousin five times removed! :smile:

> Having the uber-soft assertion type in assertj-core works because there is one API package. We have many, even in the assertj.framework bundle.

I think this is back-to-front. It's not a matter of "they're in the same package, therefore they're related" - rather, "they're related, therefore they've been placed in the same package".

Classes have their own intrinsic degrees of relatedness. This relatedness may or may not be reflected in the package hierarchies. Obviously as programmers we try and align the package hierarchies as best as we can with the classes' own intrinsic relatedness, but you would well know from experience that this is not an exact science as "relatedness" is difficult to quantify. It is possible both to end up with unrelated classes together in the same package, and also possible to have related classes in separate packages.

In the case of AssertJ core vs osgi-assertj, the decision to go with a single api package vs subpackages does not, I think, necessarily reflect differences in the underlying relatedness of the assertion classes. To put it more concretely, I don't think that (eg) `MapAssert` and `ListAssert` are intrinsically more related than (say) `BundleAssert` and `BundleContextAssert` just because the former pair are in the same package whereas the latter pair are in separate subpackages. I suspect that if AssertJ had their time over again they might not lump everything together into a single api package but would make greater use of subpackages.

> I like the add of the missing individual soft assertion providers types here.

Yes, I will keep this. I originally left them out as I anticipated an uber-type which I thought would make the individual providers redundant. But if we're not going to add an uber-type, then adding the individual providers is more important.

> But I am not sure we need to provide an uber-soft assertion provider type. People can easily make their own mixing in the things they care about.

I grant that it is not that difficult to mix your own. I guess it's a matter of convenience - if our users are frequently mixing in the same (or mostly the same) set of providers, it makes sense to provide a pre-mixed version. I thought that following the natural grouping of the original packages would be a good approximation; however, I must admit I've not had to use these personally so I'm just guessing what people might be frequently mixing in. I'm happy to leave it out for now and see where experience might take us rather than guessing.

kriegfrj:
This PR will not proceed for now. Refer to #430 for an alternative that takes care of the missing providers only.

Issue #462, opened 2022-02-20
Does not support JUnit 5's @TestInstance


chrisr3:
I have recently discovered that JUnit 5 allow you to control whether or not it reuses an instance of a test class for all tests, or creates a new instance for each test. However, the `org.osgi:org.osgi.test.junit5` extension artifact doesn’t appear to support `@TestInstance` :cry:.

This test class works:
```kotlin
package com.example.stuff.testing

import org.junit.jupiter.api.Assertions.assertEquals
import org.junit.jupiter.api.BeforeAll
import org.junit.jupiter.api.Test
import org.junit.jupiter.api.extension.ExtendWith
import org.osgi.framework.Bundle
import org.osgi.framework.BundleContext
import org.osgi.service.component.annotations.RequireServiceComponentRuntime
import org.osgi.test.common.annotation.InjectBundleContext
import org.osgi.test.junit5.context.BundleContextExtension

@ExtendWith(value = [ BundleContextExtension::class ])
@RequireServiceComponentRuntime
class WorkingTest {
    companion object {
        @InjectBundleContext
        lateinit var bundleContext: BundleContext

        private lateinit var testBundle: Bundle

        @Suppress("unused")
        @BeforeAll
        @JvmStatic
        fun setup() {
            testBundle = bundleContext.bundle
        }
    }

    @Test
    fun myTest() {
        assertEquals("testing", testBundle.symbolicName)
    }
}
```
whereas this one does not:
```kotlin
package com.example.stuff.testing

import org.junit.jupiter.api.Assertions.assertEquals
import org.junit.jupiter.api.BeforeAll
import org.junit.jupiter.api.Test
import org.junit.jupiter.api.TestInstance
import org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS
import org.junit.jupiter.api.extension.ExtendWith
import org.osgi.framework.Bundle
import org.osgi.framework.BundleContext
import org.osgi.service.component.annotations.RequireServiceComponentRuntime
import org.osgi.test.common.annotation.InjectBundleContext
import org.osgi.test.junit5.context.BundleContextExtension

@ExtendWith(value = [ BundleContextExtension::class ])
@RequireServiceComponentRuntime
@TestInstance(PER_CLASS)
class NotWorkingTest {
    @InjectBundleContext
    lateinit var bundleContext: BundleContext

    private lateinit var testBundle: Bundle

    @BeforeAll
    fun setup() {
        testBundle = bundleContext.bundle
[...]

Issue #201, opened 2020-09-09
[junit5] Better PER_CLASS lifecycle support for Jupiter extensions


kriegfrj:
*Following up from https://github.com/osgi/osgi-test/pull/181#discussion_r469706161*:

In Jupiter, there are two lifecycle modes for test classes - one new instance is created per test *method* (`PER_METHOD`), or one is created and re-used for all test methods (`PER_CLASS`). `PER_METHOD` is the default and historically what we're all used to from JUnit 4. But there are occasions where `PER_CLASS` is useful. In particular, `PER_CLASS` allows you to use non-static `@BeforeAll` and `@AfterAll` methods, which are useful in situations where you can't use static methods (eg in Kotlin, or for non-static nested test classes).

We have not paid close attention to the implementation of `PER_CLASS` test lifecycle support in our Jupiter extensions, and in particular to the combination of `PER_CLASS` lifecycle with `CONCURRENT` execution (which requires special care). For my part, this has mainly been due to a lack of understanding of `PER_CLASS` - what it does, and what it's for. But I've just now been through the exercise of implementing field injection for AssertJ's `SoftAssertionsExtension` (see joel-costigliola/assertj-core#1984) and I think I've hit upon a pattern of implementation for injector-type extensions that we should consider replicating in the osgi-test extensions.

As noted, one of the main concerns with `PER_CLASS` semantics is the state is shared across the test instance. For the most part in our extensions this is ok as we set/reset the state on every test. The main problem is if you're trying to run `PER_CLASS` and `CONCURRENT`, where the tests may be trying to access the same state concurrently.

The way around this is to use `ThreadLocal` storage for the injected fields, so that each thread accesses its own copy. This fixes the concurrency problem, but it requires a little more work on the part of the extension developer (and in some cases the test developer/extension user) to support it. In `SoftAssertionsExtension`, I figured out that when you're injecting a type that you've got some control over, it can be possible to inject a delegating implementation in this case, which hides the thread-localisation from the user. In this pattern, you initialise the field with a thread-local-aware delegator when the instance is created, and then when the test is actually run (in `beforeEach()`) you configure the delegator with the per-test instance.

Of the types that we're currently injecting, we have:
* `BundleContextExtension`: `BundleContext`, `InstallBundle`
* `ServiceExtension`: `ServiceAware`, `List<?>`, and arbitrary service types.
* `ConfigurationExtension`: `Configuration`

[...]

PR #466, opened 2022-02-23
[junit5] Populate fields for beforeAll() if PER_CLASS

kriegfrj:
This is a prototype of the fix for `BundleContextExtension` for #201, #462.

If we are happy with the approach, I will add it to the other extensions.

@chrisr3, can you take it for a spin and see if it solves your problem to your satisfaction?

### DIFF
--- a/org.osgi.test.junit5/pom.xml
+++ b/org.osgi.test.junit5/pom.xml
@@ -138,6 +138,7 @@
 									afterEach.jar,\
 									afterAll.jar,\
 									afterClass.jar
+								Bundle-Activator: org.osgi.test.junit5.test.activator.TestActivator
 								Test-Cases: ${classes;HIERARCHY_INDIRECTLY_ANNOTATED;org.junit.platform.commons.annotation.Testable;CONCRETE;PUBLIC}
 							]]></bnd>
 							<includeClassesDir>false</includeClassesDir>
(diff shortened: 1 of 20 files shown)


chrisr3:
> @chrisr3, can you take it for a spin and see if it solves your problem to your satisfaction?

My test-case now passes, and so does a larger OSGi-using project that I have. So this looks good to me :+1:.

kriegfrj:
> > @chrisr3, can you take it for a spin and see if it solves your problem to your satisfaction?
> 
> My test-case now passes, and it doesn't break a larger OSGi-using project that I have. So this looks good to me 👍.

Thanks so much for confirming - I'll try and polish it up for merging tomorrow.

bjhargrave:
@kriegfrj Can you finish this PR up? I think we need to proceed to kick off a 1.1 release.

kriegfrj:
I will try to finish this today.

kriegfrj:
Found a couple of unrelated bugs in a couple of extensions while working on this. Will fix these in a separate PR. Should have this done in a couple of hours (would be good to squeeze them into 1.1.0 too).

bjhargrave (approved):

bjhargrave:
Very nice!

kriegfrj:
> Very nice!

:+1: Thanks for the tidy-ups!

PR #482, opened 2022-03-07
Injection refactor

bjhargrave:
Refactor InjectingExtension to handle basic type validation.

For most uses, the base InjectingExtension class can validate the
target injection site type and remove that burden from subclasses.

So we allow subclasses to supply target types which will be used to
validate the injection site is compatible with at least one target
type.

### DIFF
--- a/org.osgi.test.common/src/main/java/org/osgi/test/common/inject/TargetType.java
+++ b/org.osgi.test.common/src/main/java/org/osgi/test/common/inject/TargetType.java
@@ -18,11 +18,12 @@
 
 package org.osgi.test.common.inject;
 
+import static java.util.Objects.requireNonNull;
+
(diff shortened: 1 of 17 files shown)


kriegfrj:
:+1: This was on the to-do list. `ServiceExtension` was the odd-one-out here because the others all used a fixed type. I was contemplating creating a subclass of `InjectingExtension` that the others could inherit from which does what you've done, and then have `ServiceExtension` extend directly from `InjectingExtension`. But I think your approach will work well.

A possible further improvement: There is still some duplication between handling of parameters and handling of fields. Perhaps instead of factoring out `supportsField()` and `supportsParameter()`, we need to factor out a `supportsType()` implementation which they both delegate to. Then subclasses (like `ServiceExtension`) only need to override it in one place.

PR #493, opened 2022-03-24
Adds a bnd workspace library

juergen-albert:
With the new bnd libraries, we can now offer such a functionality.

If it is acceptable, the bnd example might become smaller and easier to maintain as well. 
The library would replace most of the artifacts in the central.mvn and will always be in sync with the latest references version. If all Works out, we can also add another library, which comes with this one. With it, we can replace the test.bnd with a library for projects and even bndruns. The user in the end will only need to add `-library: osgi-test`  to his test project or bndrun and all the dependencies and settings will be taken care of.

Signed-off-by: Juergen Albert <j.albert@data-in-motion.biz>


bjhargrave (commented):
This seems like it would be better from the Bnd project. I think I would like to keep this project focused on the testing artifacts. I am personally not keen on the Bnd workspace model these days anyway.

Even if we did want to put something like this in this repo, we would need testing to ensure it actually works as we continue development in the repo. I have no way to know this even works.

Also, with the existing bom, couldn't you just use the Bnd pom repository to get all the necessary osgi-test artifacts instead of a convoluted library artifact?

juergen-albert:
> This seems like it would be better from the Bnd project. I think I would like to keep this project focused on the testing artifacts. I am personally not keen on the Bnd workspace model these days anyway.

I know that you are working more towards abending the bnd workspace for you, but the library is basically the bnd Version of the bom. It provides a simple method of adding dependencies to your workspace. I could provide this somewhere else, but then it would become out of sync the moment, I miss a new release. 
The way it is build, it will keep in sync automatically as long as you don't add new dependencies or change the naming of the version properties in the parent pom.

> Even if we did want to put something like this in this repo, we would need testing to ensure it actually works as we continue development in the repo. I have no way to know this even works.

I can add it to the bnd example, which needs to be maintained as well at the moment. With the library, this would become easier as for testing only the version of the library needs to be updated in the maven index file. Not sure how this can be automated though, as the examples and the repo is not really connected.  

> Also, with the existing bom, couldn't you just use the Bnd pom repository to get all the necessary osgi-test artifacts instead of a convoluted library artifact?

True, but the BndPomRepository is evil and we try to get rid of it where ever we can. @stbischof and I would gladly have an eye on the library in regard to testing it before a new release.

Issue #496, opened 2022-03-28
[InjectingExtension] Refactoring suggestions


kriegfrj:
Having reviewed the `InjectingExtension` in more detail, I believe that we've made some mistakes in the pattern that we should consider addressing:

1. It became apparent in my work on #495 and #494 that `supportsType()` should have taken the resolved annotation as a parameter. We can add this as a new method and delegate to the old.
2. `InjectingExtension` implements `AfterEachCallback` and `AfterAllCallback`. However, the implementations do nothing. There is therefore no reason for the base class to implement these methods - if a specific implementation wishes to add this functionality, it can do so by directly implementing them.
3. There is no reason for `supportsType()` to return a boolean. `supportsParameter()`and `supportsField()` both do a check to see if the field/parameter is annotated by the annotation type that is "owned" by the extension. If it is not owned by the extension, then it should return <tt>false</tt> so that Jupiter can find the rightful owner. But if if it is owned by the extension, then supportsField()/supportsParameter() must either return <tt>true</tt> (if all is well) or else throw an informative exception if there is some kind of misconfiguration. Simply returning <tt>false</tt> at this stage would be uninformative and implementations should be discouraged from doing so. `boolean supportsType()` should have perhaps been `void checkField()`.

Were it not for the fact that `InjectingExtension` has already found its way into a release in its current form, I would simply make these changes. However, I'm looking at the versioning consequences.

The first change would introduce a minor or a major change depending on whether we simply change the signature or else add a new method that delegates to the old. The second would be a major change. The third - we could simply ignore the boolean return value from `supportsType()` and avoid any formal change, though the practical change would still be there so it might be misleading to *not* bump the version in that case (I think it would be confusing to implementers have a boolean-valued override whose return value is ignored).

Thoughts?

bjhargrave:
> 1. that `supportsType()` should have taken the resolved annotation as a parameter. We can add this as a new method and delegate to the old.

I am not sure what useful information the annotation has for supportsType? The TargetType parameter is the proper place (abstraction) to pass any necessary information to the method. If we need more information, we should improve TargetType and use it to pass information.

kriegfrj:
> > 2. `InjectingExtension` implements `AfterEachCallback` and `AfterAllCallback`. However, the implementations do nothing.
> 
[...]

PR #497, opened 2022-03-28
[ServiceExtension] Better sanity checking and wildcard support

kriegfrj:
I'm expecting this build to fail with a versioning error.

The versioning error is easily fixed once a decision is made in respect to #496.

### DIFF
--- a/org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceExtension.java
+++ b/org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceExtension.java
@@ -64,6 +64,11 @@ public class ServiceExtension extends InjectingExtension<InjectService> {
 	@Override
 	protected boolean supportsType(TargetType targetType, ExtensionContext extensionContext)
 		throws ParameterResolutionException {
+		extractValueType(targetType);
+		return true;
+	}
+
+	private Class<?> extractValueType(TargetType targetType) {
 		Type serviceType;
 		if (targetType.matches(List.class) || targetType.matches(ServiceAware.class)) {
 			if (targetType.hasParameterizedTypes()) {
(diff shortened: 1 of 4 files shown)


org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceExtension.java
```diff
@@ -62,57 +62,73 @@ public ServiceExtension() {
 	}
 
 	@Override
-	protected boolean supportsType(TargetType targetType, ExtensionContext extensionContext)
+	protected boolean supportsType(TargetType targetType, InjectService injectService,
```
bjhargrave, line 65: This is wrong. Any information about the injection site should be in the TargetType parameter. Passing the annotation is leaky. Improve the information in TargetType if it does not already contain what you need.

bjhargrave (changes requested):

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceExtension.java
kriegfrj: `ServiceExtension` was the special case here because whether or not it supports a type depends on whether or not the `service` attribute has been set on the annotation.

I managed to work around it by instead moving the validation code from `supportsType()` into `resolveValue()` (which already has access to the annotation instance). This had the bonus of getting rid of the duplicated validation code and not needing the new method which required a version bump.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/service/ServiceExtension.java
```diff
@@ -62,57 +62,71 @@ public ServiceExtension() {
...
 	@Override
-	protected boolean supportsType(TargetType targetType, ExtensionContext extensionContext)
+	protected boolean supportsType(TargetType targetType, ExtensionContext extensionContext) {
+		return true;
+	}
+
+	private Type extractServiceType(TargetType targetType, InjectService injectService)
 		throws ParameterResolutionException {
-		Type serviceType;
+		Type valueType, upperBoundType = null;
```
bjhargrave, line 71: I don't care for this style of variable declaration. It is non-obvious since you declare multiple variables and initialize one(?). Better to have one variable declaration per line.

bjhargrave (changes requested):
I am ok with this given my one comment on variable declaration.

PR #500, opened 2022-03-28
[BundleContextExtension] More informative error on null BundleContext

### LINKED ISSUE #302: BundleContextExtension does not check for null
If the bundles are not started that are used to acquire the bundle-context then null is sued internally leading to NPE in BundleContextExtension.

kriegfrj:
Fixes #302.


rotty3000 (approved):
LGTM

PR #502, opened 2022-03-28
extract `getURL` in BundleInstaller.EmbeddedLocation

stbischof:
allows users to have a easy way to get the URL of a nested jar.


org.osgi.test.common/src/main/java/org/osgi/test/common/install/BundleInstaller.java
```diff
@@ -207,6 +207,12 @@ public int hashCode() {
 
 		public InputStream openStream(BundleContext bc) throws IOException {
 
+			URL jarEmbeddedFileUrl = getURL(bc);
+
+			return jarEmbeddedFileUrl.openStream();
+		}
+
+		public URL getURL(BundleContext bc) throws FileNotFoundException {
```
bjhargrave, line 215: I think `IOException` is a better choice for the throws.

bjhargrave (changes requested):

bjhargrave:
Also, the build broke due to a baseline error.

org.osgi.test.common/src/main/java/org/osgi/test/common/install/BundleInstaller.java
stbischof: it is.

org.osgi.test.common/src/main/java/org/osgi/test/common/install/BundleInstaller.java
```diff
@@ -223,10 +228,9 @@ public InputStream openStream(BundleContext bc) throws IOException {
 			URL jarEmbeddedFileUrl = entryStream
 				.findFirst()
 				.orElseThrow(
-					() -> new FileNotFoundException(String.format("File [%s/%s] not found in Bundle [%s:%s]", path,
+					() -> new IOException(String.format("File [%s/%s] not found in Bundle [%s:%s]", path,
```
bjhargrave, line 231: You can still throw FileNotFoundException here. It is a subclass of IOException. My point was about the method declaration's throws clause and not the actual exception thrown.

bjhargrave (approved):

PR #503, opened 2022-03-29
Add an example that could be used inside PDE

laeubi:
This is an example where one can run osgi-test from inside [Eclipse PDE](https://www.eclipse.org/pde/)

To execute it do the following:

- make sure you are using a recent Eclipse for RCP/RAP Developers edition (minimum 2022-03)
- open the 'example.target' and set it as your active target
- import the projects, right click on PlayerTest.java and choose `Run As` > `Junit Plugin Test`

### DIFF
--- a/examples/osgi-test-example-pde/README.md
+++ b/examples/osgi-test-example-pde/README.md
@@ -1,0 +1,8 @@
+# osgi-test-example-pde
+An example project demonstrating OSGi testing capabilities using [Eclipse PDE](https://www.eclipse.org/pde/).
+
+To execute it do the following:
+
+- make sure you are using a recent Eclipse for RCP/RAP Developers edition (minimum 2022-03)
+- open the 'example.target' and set it as your active target
+- import the projects, right click on PlayerTest.java and choose `Run As` > `Junit Plugin Test`
--- a/examples/osgi-test-example-pde/bundles/org.osgi.test.example.api/.classpath
+++ b/examples/osgi-test-example-pde/bundles/org.osgi.test.example.api/.classpath
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="src" path="src"/>
+	<classpathentry kind="output" path="bin"/>
+</classpath>
--- a/examples/osgi-test-example-pde/bundles/org.osgi.test.example.api/.project
+++ b/examples/osgi-test-example-pde/bundles/org.osgi.test.example.api/.project
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>example.api</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.jdt.core.javabuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.ManifestBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.pde.SchemaBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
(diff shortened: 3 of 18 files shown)


bjhargrave (commented):
How do we test this in our CI builds? The other examples focus on command line builds (gradle, maven). This PR includes classpath files which are brittle (they require JDK 11). 

This would be more interesting if it were a maven tycho build which could be verified by CI and we can keep up to date. I am not interested in this in its current state.

laeubi:
> How do we test this in our CI builds?

It is an example for usage in the with the Eclipse PDE tooling and not a commandline / ci.

> This PR includes classpath files which are brittle (they require JDK 11).

Whats wrong with JDK 11? Other examples require bnd, gradle ...

> I am not interested in this in its current state.

then probably you should in the next [eclipse-con presentation](https://www.eclipsecon.org/sites/default/files/slides/OSGi_Test_EclipseCon2020_Presentation.pdf) not state that "osgi-test should work with PDE (note: this has not been tested by us)" and instead write "but we are not interested and don't care" ... 

> if it were a maven tycho build which could be verified by CI

Tycho can build it but has nothing to do with PDE as tycho does not use PDE... but feel free to came up with a [PDE build](https://www.eclipse.org/pde/pde-build/) I just have no clue how to use it and don't know anyone still using it...

bjhargrave:
> It is an example for usage in the with the Eclipse PDE tooling and not a commandline / ci.

Yes, I understand that. This is why I don't think it is a good idea as we cannot continuously test to make sure it still works.


> Whats wrong with JDK 11? Other examples require bnd, gradle ...

Nothing per se, but the example does not work unless the user has JDK 11 installed into their Eclipse configuration. And when they don't, they get a lot red errors in Eclipse and no real explanation as to why.

> then probably you should in the next [eclipse-con presentation](https://www.eclipsecon.org/sites/default/files/slides/OSGi_Test_EclipseCon2020_Presentation.pdf) not state that "osgi-test should work with PDE (note: this has not been tested by us)" and instead write "but we are not interested and don't care" ...

It is not that we do not care, but if you ship something, you should have confidence it works. With no way to test, on a continuous basis, that the example works, the example will eventually cease to be valid. For example, I have no idea that the example in the PR even works right now. It could be broken. I would need to check out this PR, install Eclipse PDE and try to load it (I hope I have JDK 11 configured into that Eclipse installation...). With a CI build, this PR would be able to build and test the example to confirm it does in fact work.

> Tycho can build it but has nothing to do with PDE as tycho does not use PDE... but feel free to came up with a [PDE build](https://www.eclipse.org/pde/pde-build/) I just have no clue how to use it and don't know anyone still using it...

I know that Tycho is used to build Eclipse which can also be built from the PDE information. So there must be some relationship. I don't know Tycho either nor do I use PDE. But I do think we must be able to build the code in the repo in our CI build.

laeubi:
> Nothing per se, but the example does not work unless the user has JDK 11 installed into their Eclipse configuration

Recent eclipse needs at least Java 11 and the classpath reference only the **profile** JavaSE-11 that could be mapped to any JVM with version 11+ so you only run into problem when you remove all but a JDK8 ... that's very uncommon setup I think...

> This is why I don't think it is a good idea as we cannot continuously test to make sure it still works

It still is useful to prove it does work in an PDE also, if it might fail in the future one probably has to adjust it, but having no (manual) test case at all is even worse I think... but in the end its of course the choice of the project, we can simply keep this PR open just in case someone asks for an (continuous untested) example.

> So there must be some relationship

Yes but tycho only *uses* the information from PDE but has completely different code when it comes to executing tests (most of that parts are pluggable and require a running UI in PDE/Eclipse).

> But I do think we must be able to build the code in the repo in our CI build.

Integrate this in a CI build would require a lot of work (e.g xvnc server, complete eclipse product launch, probably something like [RCPTT](https://www.eclipse.org/rcptt/) and that's beyond what I like to invest here at the moment given that it probably would still not be accepted by (to me) unknown project rules.

bjhargrave:
@laeubi Would it be possible to update this example to be a Maven/Tycho example? Then (I assume), the example could be used in PDE and we can also run the maven build in the CI build to confirm the example is correct.

laeubi:
I'll give it a try, it just should be taken into account that PDE does not use Tycho so actually if it works on Tycho ist not a proof it will work in the PDE as well but I think thats probably acceptable.

juergen-albert:
@laeubi are you plan on doing something here? This might be a good point for an example with the new bnd pde hybrid?

laeubi:
I plan to improve OSGI-Testing support in PDE, but if you aim for `pde.bnd` usage you should be able to use the usual bnd test classpath and execute it with either bnd-maven-plugin or use https://tycho.eclipseprojects.io/doc/latest/TestingBundles.html#bnd-testing

Issue #541, opened 2022-08-01
Provide a Junit5 Extension to start an OSGi-Connect-Framework as part of the test without any need for special setup


laeubi:
### Current state

OSGi-Test provides a very convenient and powerful way to (unit) test OSGi code, but the current setup is rather complex and hard to understand for anyone not familiar with OSGi+BND tooling, e.g. [the maven example](https://github.com/osgi/osgi-test/tree/main/examples/osgi-test-example-mvn) requires the following stuff:

1. A non trivial test.bndrun file https://github.com/osgi/osgi-test/blob/bf9b049afb4bf3331cfa5a222d7566f7bc3ae29f/org.osgi.test.common/test.bndrun#L1-L62
2. `bnd-resolver-maven-plugin` with non-trivial configuration https://github.com/osgi/osgi-test/blob/bf9b049afb4bf3331cfa5a222d7566f7bc3ae29f/examples/osgi-test-example-mvn/pom.xml#L134-L185
3. `bnd-testing-maven-plugin` with non trivial configuration https://github.com/osgi/osgi-test/blob/bf9b049afb4bf3331cfa5a222d7566f7bc3ae29f/examples/osgi-test-example-mvn/pom.xml#L162-L185
4. A disabling of the `maven-surefire-plugin` https://github.com/osgi/osgi-test/blob/bf9b049afb4bf3331cfa5a222d7566f7bc3ae29f/examples/osgi-test-example-mvn/org.osgi.test.example.player.test/pom.xml#L91-L96
5. `bnd-maven-plugin` with non trivial configuration for the test-plugin itself https://github.com/osgi/osgi-test/blob/bf9b049afb4bf3331cfa5a222d7566f7bc3ae29f/examples/osgi-test-example-mvn/org.osgi.test.example.player.test/pom.xml#L100-L113

### Problem with current aproach
While all of this is flexible and powerful in a larger OSGi application setup, it is simply an integration barrier for many other smaller use-cases

1. I have a library project and I don't know much about OSGi but still people ask me to add OSGi support (e.g. https://github.com/xerial/sqlite-jdbc/pull/738) for a very limited part of OSGi. Suggesting such a complex setup that requires to restructure large parts of the project will most likely be rejected.
2. I don't want to use an extra project for test but simply use Surefire plugin as-is, and I don't need OSGi for every test so I only selectively want to enable osgi-test it for some test but leave the others alone
[...]

PR #544, opened 2022-08-04
Junit5 Extension to start an OSGi-Connect-Framework as part of the test

laeubi:
The current setup is rather complex and hard to understand for anyone not familiar with OSGi+BND tooling.

This adds a JUnit5 Extension on top of the OSGi Framework connect that drastically reduces the complexity but still enables usage of osgi-test in a unit test environment.

Fixes https://github.com/osgi/osgi-test/issues/541


laeubi:
@rotty3000 can you please enable the workflow so I can see if there are any build-issues with my contribution?

kriegfrj:
> @rotty3000 can you please enable the workflow so I can see if there are any build-issues with my contribution?

Done! Sorry, didn't realise this was waiting for a confirmation...

laeubi:
> Done! Sorry, didn't realise this was waiting for a confirmation...

This needs manual confirmation for first-time contributors https://github.blog/changelog/2021-04-22-github-actions-maintainers-must-approve-first-time-contributor-workflow-runs/

kriegfrj:
> > Done! Sorry, didn't realise this was waiting for a confirmation...
> 
> This needs manual confirmation for first-time contributors https://github.blog/changelog/2021-04-22-github-actions-maintainers-must-approve-first-time-contributor-workflow-runs/

Yes, the message was very helpful in explaining that to me, and it makes sense. I just wasn't aware that intervention was needed.

This project sounds interesting, and a sincere thankyou for making the effort to . I'm terribly sorry that I don't have the bandwidth to take a closer look at it myself (besides, I'm not much of an expert at OSGi Connect). Hopefully someone else will get a chance to have a look soon.

laeubi:
@kriegfrj I don't think you need to be an connect expert to review this, as for the usual user it is quite irrelevant what is the underlying technique. So I think you can perfectly review this e.g. is the readme enough to use that feature or do it misses important instructions, is the provided example complete enough to prove the function itself, do you think that the code fits into osgi-test and so on, any input is welcome here that helps getting this feature into a release :-)

Of course one can think of a lot more features and/or enhancements that could then better be provided as separate (smaller) PRs.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/framework/FrameworkExtension.java
```diff
@@ -0,0 +1,277 @@
...
+			case Bundle.RESOLVED :
+				return "RESOLVED ";
+			case Bundle.STARTING :
+				return "STARTING ";
+			case Bundle.STOPPING :
+				return "STOPPING ";
+			default :
+				return String.valueOf(state);
+		}
+	}
```
kriegfrj, line 275: Have a look at https://github.com/osgi/osgi-test/tree/main/org.osgi.test.common/src/main/java/org/osgi/test/common/bitmaps for a possible alternative that might avoid repeating yourself on this stuff. There is a bitmap defined for BundleState, perhaps a ComponentState one would be good too.

kriegfrj:
> @kriegfrj I don't think you need to be an connect expert to review this, as for the usual user it is quite irrelevant what is the underlying technique. So I think you can perfectly review this e.g. is the readme enough to use that feature or do it misses important instructions, is the provided example complete enough to prove the function itself, do you think that the code fits into osgi-test and so on, any input is welcome here that helps getting this feature into a release :-)
>

I'm certain I could review some of it. But also (and I am terribly sorry) I have some other competing time pressures that do not allow me to spend the time looking at this that it deserves and requires. (This is not a reflection on the quality of your work. :smile:)

I think that the concept is sound. I think that it is arguably a fit for osgi-test. I can see similarities to Launchpad (which is used in bnd core). I have made one suggestion that jumped out at me in my superficial review. I wish I could do more.

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/framework/FrameworkExtension.java
laeubi: Done, thanks for the hint, I think it would actually be good to have this directly in the OSGi framework :-)

bjhargrave:
Some high level comments:

This seems very overlapping with [LaunchPad](https://github.com/bndtools/bnd/tree/master/biz.aQute.launchpad/test/aQute/launchpad) in Bnd. LaunchPad does not use OSGi Connect but launches a child framework instance. We should really not promote two competing ways to do basically the same thing. Ideally, we should harmonize here.

Not all bundles can run in OSGi Connect as it does not provide class space isolation. This is why LaunchPad launches a child framework. So if we include OSGi Connect support in this project, it needs to be in its own `org.osgi.test.junit5.connect` project separate from the "normal" OSGi test support. This also does not require raising the OSGi framework version for other projects which should continue to work on older OSGi framework versions.

OSGi Connect support also needs to work with other OSGi frameworks like Felix. I think I saw some Equinox specifics in the code.

New annotations to support OSGi Connect should not be in the `org.osgi.test.common.annotation` package. That package is for annotations whose behavior is "shared" with the JUnit 4 support. So the new annotations should be in `org.osgi.test.junit5.connect.annotation` (or similar) in the new `org.osgi.test.junit5.connect` project.

laeubi:
> This seems very overlapping with [LaunchPad](https://github.com/bndtools/bnd/tree/master/biz.aQute.launchpad/test/aQute/launchpad) in Bnd

Mi first priority here is to not introduce any bnd- requirements (bnd files, bnd workspace, bnd resolve... whatever) but run from a  "whatever assemble this" classpath, so if Launchpad can be used in such a way it would be great to show an example like provided with this PR:

1. a single API bundle
2. a single Impl bundle including the impl and the test

that would be helpful to decide if it could be harmonized. Beside that, if one wants could even have an embedded "connect" and "launchpad" variant and user can choose if we make the annotations more generic, that will only require a slightly different setup

> Not all bundles can run in OSGi Connect as it does not provide class space isolation.

That's wrong OSGi connect can do anything as a "regular" OSGi Framework including class space isolation. I even added an option to let the user choose if a bundle should be isolated see
https://github.com/osgi/osgi-test/blob/2fe1b6092b0e9fa36cb3217518de81d4f3be438a/org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/framework/JUnit5Module.java#L99-L104

> OSGi Connect support also needs to work with other OSGi frameworks like Felix

This will work with any OSGi Connect compliant framework, no idea if Felix implements OSGi Connect... is there some overview about all frameworks?

laeubi:
In the meanwhile I enhanced this and deployed a first working version to maven central:

https://github.com/laeubisoft/osgi-test-framework

just let me know if there is interest in moving the code over here ... I think this would greatly simplify tests for "mixed setups" and has helped me already in some projects.

PR #595, opened 2022-12-21
junit5: Add cardinality element to ServiceSource

bjhargrave:
This allows the user to specify the desired cardinality (like InjectService). The default cardinality is now 1. It was previously 0 but that seems wrong.

If the cardinality is not met, a ParameterResolutionException is thrown with an error message.


rotty3000:
The default cardinality of 0 for ServiceAware was for cases where the service is published by the test itself. but as long as we can reset the cardinality to 0 so that there isn't a wait is fine.

bjhargrave:
> The default cardinality of 0 for ServiceAware

This is about ServiceSource for parameterized test methods. One can still specify `cardinality=0` but I am not sure they is even practical since JUnit 5 may complain that there is zero parameterization for a test method.

stbischof:
thank you @bjhargrave

PR #744, opened 2023-11-06
Update Jupiter for Eclipse 09-23

maho7791:
- Updated Junit Jupiter to work with Eclipse 09-23 IDE
- use bnd 7.0.0 to build

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,18 +29,18 @@
 	<version>${revision}</version>
 
 	<properties>
-		<revision>1.3.0-SNAPSHOT</revision>
+		<revision>1.4.0-SNAPSHOT</revision>
 		<project.build.outputTimestamp>1980-02-01T00:00:00Z</project.build.outputTimestamp>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<java.version>1.8</java.version>
 		<java.release>8</java.release>
 		<maven.compiler.source>${java.version}</maven.compiler.source>
 		<maven.compiler.target>${java.version}</maven.compiler.target>
-		<bnd.version>6.4.0</bnd.version>
+		<bnd.version>7.0.0</bnd.version>
 		<junit-jupiter.compile.version>5.6.0</junit-jupiter.compile.version>
 		<junit-platform.compile.version>1.6.0</junit-platform.compile.version>
-		<junit-jupiter.version>5.9.2</junit-jupiter.version>
-		<junit-platform.version>1.9.2</junit-platform.version>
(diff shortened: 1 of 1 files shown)


kriegfrj:
Hey @maho7791 , thanks for the PR! What's the end goal here, if you don't mind me asking? I'm surprised that osgi-test doesn't already work with Eclipse 09-23?

maho7791:
The latest JUnit Laucher in the PDE 09-23 expects Jupiter 5.10. to work. If OSGi Test uses a dependency lower than 5.10 you end up MethodNotFound exceptio. This PR is just for completeness, so that OSGi Test also works with 5.10. 
If I think about it, should we consider also increase the jupiter and platform-compile-version?

kriegfrj:
> The latest JUnit Laucher in the PDE 09-23 expects Jupiter 5.10. to work. If OSGi Test uses a dependency lower than 5.10 you end up MethodNotFound exceptio. 

If it's not too difficult, can you put the stack trace of that exception in here somewhere for posterity?

> This PR is just for completeness, so that OSGi Test also works with 5.10. 

I think that the problem is not that OSGi Test won't work with 5.10 - I think it will as the JUnit team are pretty good about maintaining backward compatibility. The problem seems to be that the *Eclipse 09-23's JUnit integration* requires JUnit Platform >=1.10. So you can't run the tests inside the IDE. I'm guessing that they probably still work from Maven command line though,

> If I think about it, should we consider also increase the jupiter and platform-compile-version?

I am not against increasing the baseline version that we compile against if there is a specific need for a newer feature in the newer version. However, in the absence of such a need I am hesitant.

So if changing the runtime version fixes your problem with the IDE integration, perhaps we should stick with that for this PR. We can always revisit bumping up the compile version at some point if/when there is a case to be made to use a newer feature.

pom.xml
```diff
@@ -29,7 +29,7 @@
 	<version>${revision}</version>
 
 	<properties>
-		<revision>1.3.0-SNAPSHOT</revision>
+		<revision>1.4.0-SNAPSHOT</revision>
```
kriegfrj, line 32: 1.3.0 has not been released yet, so there is no need to bump this to 1.4.0.

kriegfrj (changes requested):
Thanks for this!

Given that the second commit simply undoes one of the changes in the first commit, can we also squash these two into a single commit before merging?

timothyjward:
@maho7791 - is this still a problem, or can we close this PR?

maho7791:
not needed

PR #777, opened 2024-01-18
[junit5.cm] handle just configurations from annotations

stbischof:
Fix: handle just configurations from annotations

@timothyjward the pom structure has one version for all bundles. should we switch to independent bundleversions. defind in parent-poms properties?


timothyjward:
>@timothyjward the pom structure has one version for all bundles. should we switch to independent bundleversions. defind in parent-poms properties?

That depends - if we think that we might release them independently then yes, but I'm not sure whether that is currently possible?

timothyjward:
We should definitely add some new tests for the desired behaviour.

stbischof:
would be great is you could add the tests

PR #779, opened 2024-01-23
Add tests to show that external configurations are not perturbed

timothyjward:
Previous versions of the test infrastructure would cause spurious updates in configurations provided from other management agents. This test aims to validate that this no longer happens in an effort to avoid regressions.


org.osgi.test.junit5.cm/src/test/java/org/osgi/test/junit5/cm/test/lifecycle/ExternalConfigLifecycleTest.java
```diff
@@ -0,0 +1,157 @@
...
+	private static long			inititalChangeCount;
+
+	private static int			updateCount	= 0;
+
+	@BeforeAll
+	static void beforeAll(@InjectConfiguration(CONFIG)
+	Configuration cm) throws IOException {
+		inititalChangeCount = cm.getChangeCount();
+		updateCount = 0;
+		checkExternalConfig(cm, 0);
```
stbischof, line 56: use `updateCount` not `0` in `checkExternalConfig(cm, updateCount);`

PR #786, opened 2024-02-05
Annotations with location NOT_SET should use null locations

### LINKED ISSUE #615: [cm] either create unbound config, or provide an option for it.
Currently the annotation performs the following:

```
if (Property.NOT_SET.equals(configAnnotation.location())) {
   configuration = configurationAdmin.getFactoryConfiguration(configAnnotation.factoryPid(),
	configAnnotation.name());
} else {
   configuration = configurationAdmin.getFactoryConfiguration(configAnnotation.factoryPid(),
	configAnnotation.name(), configAnnotation.location());
}
```

that way configurations are either require an **explicit** location or are bound to the "calling bundle" (what is the test probe) and is most likely not desired by default.

This should be changed to using a `null` location, or if it is absolutely desired, at least a boolean flag to do so.

See [104.4.2 Dynamic Binding ](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.cm.html#d0e11038)

another option would be to not use `Property.NOT_SET` but `?` as the default value...

timothyjward:
When processing annotations a NOT_SET location was resulting in the configuration location being set to the tester bundle - this is clearly incorrect and should explicitly unset as per the ConfigurationAdmin API.

Fixes #615

### DIFF
--- a/org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/ConfigurationExtension.java
+++ b/org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/ConfigurationExtension.java
@@ -168,7 +168,7 @@ public class ConfigurationExtension extends InjectingExtension<InjectConfigurati
 
 			Configuration configuration;
 			if (Property.NOT_SET.equals(configAnnotation.location())) {
-				configuration = configurationAdmin.getConfiguration(configAnnotation.pid());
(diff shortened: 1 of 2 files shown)


timothyjward:
@laeubi I believe this addresses the problem that you had?

stbischof (approved):

stbischof:
Lgtm

PR #804, opened 2024-02-13
Revert "build: Bump ch.qos.logback:logback-classic from 1.3.4 to 1.4.14 in /examples/osgi-test-example-gradle"

bjhargrave:
Reverts osgi/osgi-test#760

Logback 1.4 is for Jakarta EE and requires Java 11.


bjhargrave:
We need to use logback 1.3.x for the gradle example.

PR #806, opened 2024-02-15
Add support for system properties in configuration annotations

### LINKED ISSUE #614: [cm] Support Systemproperties in Property
Currently one can use the following construct:

```
@WithFactoryConfiguration(factoryPid = ServerRuntime.NAME, name = "testme", properties = {
			@Property(key = "port", value = "9999") })
```
This is great unless the port should be able to be set from the outside (e.g. different test-systems might have the port already allocated, e.g. on CI or dev maschine)

It would therfore be good to support an additional property `systemProperty` in the `@Property` annotation that  evaluates the given name as a system Property e.g.:

```
@WithFactoryConfiguration(factoryPid = ServerRuntime.NAME, name = "testme", properties = {
			@Property(key = "port", value = "9999", systemProperty = "test.port") })
```

if the process is run with `-Dtest.port=1234` the the value should actually be derived as `System.getProperty("test.port", <value>)`

timothyjward:
This commit adds the systemProperty member to the @Property annotation. If set then the System Properties will be queried using that property name, and the value used to set the configuration. If there is no matching System Property for the supplied key then the @Property annotation value member will be used instead. If the @Property is not of type Scalar then the value of the System Property will be split using ',' as a separator.

Fixes #614

### DIFF
--- a/org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/PropertiesConverter.java
+++ b/org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/PropertiesConverter.java
@@ -17,10 +17,13 @@
  *******************************************************************************/
 package org.osgi.test.common.annotation;
 
+import static org.osgi.test.common.annotation.Property.NOT_SET;
+
 import java.lang.reflect.Array;
 import java.util.Arrays;
 import java.util.Dictionary;
 import java.util.Hashtable;
+import java.util.Properties;
 
 import org.osgi.test.common.annotation.Property.Scalar;
 import org.osgi.test.common.annotation.Property.Type;
@@ -39,9 +42,24 @@ public class PropertiesConverter {
 
 		boolean primitive = entry.type()
 			.equals(Type.PrimitiveArray);
-		Object result = createArray(entry.scalar(), primitive, entry.value().length);
+		String[] value;
+		if (NOT_SET.equals(entry.systemProperty())) {
+			value = entry.value();
+		} else {
+			Properties sysProps = System.getProperties();
+			if (sysProps.containsKey(entry.systemProperty())) {
+				String prop = sysProps.getProperty(entry.systemProperty());
+				value = entry.type() == Type.Scalar ? new String[] {
+					prop
+				} : prop.split(",");
+			} else {
+				value = entry.value();
+			}
+		}
+
+		Object result = createArray(entry.scalar(), primitive, value.length);
 		int i = 0;
-		for (String v : entry.value()) {
+		for (String v : value) {
 			Object val = null;
 
 			if (v != null) {
--- a/org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/PropertiesConverter.java
+++ b/org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/PropertiesConverter.java
@@ -21,8 +21,8 @@ import java.util.Dictionary;
 
 /**
  * The generic PropertiesConverter is unable to support all forms of property
- * conversion and should not be used. Use the JUnit version specific alternative
- * instead.
+ * conversion and should not be used. Use the JUnit 4 specific
+ * {@link org.osgi.test.junit4.properties.PropertiesConverter} instead.
  */
 @Deprecated
 public class PropertiesConverter {
--- a/org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/Property.java
+++ b/org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/Property.java
@@ -55,6 +55,26 @@ public @interface Property {
 		PrimitiveArray
 	}
 
+	public enum ValueSource {
+		/** Use the {@link Property#value()} directly */
+		Value,
+		/**
+		 * Use the first {@link Property#value()} as a key in the System
+		 * Properties
+		 */
+		SystemProperty,
+		/**
+		 * Use the first {@link Property#value()} as a key in the Environment
+		 */
+		EnvironmentVariable,
+		/** Use the name of the test class */
+		TestClass,
+		/** Use the name of the test method */
+		TestMethod,
+		/** Use the test unique identifier */
+		TestUniqueId
+	}
+
 	String key();
 
 	String[] value() default "";
@@ -63,6 +83,9 @@ public @interface Property {
 
 	Type type() default Type.Scalar;
 
-	String systemProperty() default NOT_SET;
+	/**
+	 * @return the source that should be used to obtain the value
+	 */
+	ValueSource source() default ValueSource.Value;
 
 }
--- a/org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/Property.java
+++ b/org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/Property.java
@@ -63,4 +63,6 @@ public @interface Property {
 
 	Type type() default Type.Scalar;
 
+	String systemProperty() default NOT_SET;
+
 }
--- a/org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/Property.java
+++ b/org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/Property.java
@@ -65,11 +65,22 @@ public @interface Property {
 		Value,
 		/**
 		 * Use the first {@link Property#value()} as a key in the System
-		 * Properties
+		 * Properties. If the key does not exist then either:
+		 * <ul>
+		 * <li>Use the second {@link Property#value()} as the value</li>
+		 * <li>Throw an exception indicating that there is no property for the
+		 * supplied key</li>
+		 * </ul>
 		 */
 		SystemProperty,
 		/**
-		 * Use the first {@link Property#value()} as a key in the Environment
+		 * Use the first {@link Property#value()} as a key in the Environment If
+		 * the key does not exist then either:
+		 * <ul>
+		 * <li>Use the second {@link Property#value()} as the value</li>
+		 * <li>Throw an exception indicating that there is no property for the
+		 * supplied variable name</li>
+		 * </ul>
 		 */
 		EnvironmentVariable,
 		/** Use the name of the test class */
--- a/org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/Property.java
+++ b/org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/Property.java
@@ -55,6 +55,11 @@ public @interface Property {
 		PrimitiveArray
 	}
 
+	/**
+	 * Indicates the source of the configuration value
+	 *
+	 * @since 1.2
+	 */
 	public enum ValueSource {
 		/** Use the {@link Property#value()} directly */
 		Value,
@@ -85,6 +90,7 @@ public @interface Property {
 
 	/**
 	 * @return the source that should be used to obtain the value
+	 * @since 1.2
 	 */
 	ValueSource source() default ValueSource.Value;
 
--- a/org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/package-info.java
+++ b/org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/package-info.java
@@ -17,7 +17,7 @@
  *******************************************************************************/
 
 @Export(attribute = "junit=4")
-@Version("1.1.0")
+@Version("1.2.0")
 package org.osgi.test.common.annotation;
 
 import org.osgi.annotation.bundle.Export;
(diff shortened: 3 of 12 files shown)


timothyjward:
@laeubi - is this sufficient to cover your use case in #614 ?

org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/PropertiesConverter.java
```diff
@@ -39,9 +42,24 @@ private static Object toValue(Property entry) {
...
 		boolean primitive = entry.type()
 			.equals(Type.PrimitiveArray);
-		Object result = createArray(entry.scalar(), primitive, entry.value().length);
+		String[] value;
+		if (NOT_SET.equals(entry.systemProperty())) {
+			value = entry.value();
+		} else {
+			Properties sysProps = System.getProperties();
+			if (sysProps.containsKey(entry.systemProperty())) {
+				String prop = sysProps.getProperty(entry.systemProperty());
```
bjhargrave, line 51:
```suggestion
				String prop = sysProps.getProperty(entry.systemProperty()).trim();
```

To elide leading and trailing whitespace.

laeubi:
@timothyjward yes this looks like it would solve that issue!

org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/Property.java
```diff
@@ -63,4 +63,6 @@ public enum Type {
 
 	Type type() default Type.Scalar;
 
+	String systemProperty() default NOT_SET;
```
bjhargrave, line 66: What if `value` and `systemProperty` are both set? The current code prefers systemProperty over value but this is not documented. 

Perhaps a better choice is to define the source of the actual value.

```suggestion
    public enum Source { String, SystemProperty, EnvironmentVariable }:
	Source source() default String;
```

`String` is the current behavior. `SystemProperty` means the value holds a system property name. `EnvironmentVariable` means the value holds an env var name.

This would allow for much more future proofing and avoid ambiguity.

stbischof:
Could we use this to also set  the TestClass TestMethod or TestUniqueId fom the executionCintest to the config properties?

timothyjward:
> Could we use this to also set the TestClass TestMethod or TestUniqueId fom the executionCintest to the config properties?

Possibly, but it's going to involve moving the PropertiesHandler as we need access to an ExecutionContext, which is only visible in JUnit5.

timothyjward:
@bjhargrave - is this more like you had in mind?

org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/PropertiesConverter.java
```diff
@@ -17,173 +17,19 @@
...
-import java.util.Arrays;
 import java.util.Dictionary;
-import java.util.Hashtable;
-
-import org.osgi.test.common.annotation.Property.Scalar;
-import org.osgi.test.common.annotation.Property.Type;
 
+/**
+ * The generic PropertiesConverter is unable to support all forms of property
+ * conversion and should not be used. Use the JUnit version specific alternative
```
bjhargrave, line 24: Since this is in the junit4 project's source folder, why not have the comment refer to the specific `org.osgi.test.junit4.properties.PropertiesConverter` replacement class?

org.osgi.test.junit5/src/main/java/org/osgi/test/common/annotation/PropertiesConverter.java
```diff
@@ -17,173 +17,19 @@
...
-import java.util.Arrays;
 import java.util.Dictionary;
-import java.util.Hashtable;
-
-import org.osgi.test.common.annotation.Property.Scalar;
-import org.osgi.test.common.annotation.Property.Type;
 
+/**
+ * The generic PropertiesConverter is unable to support all forms of property
+ * conversion and should not be used. Use the JUnit version specific alternative
```
bjhargrave, line 24: Since this is in the junit5 project's source folder, why not have the comment refer to the specific `org.osgi.test.junit5.properties.PropertiesConverter` replacement class?

org.osgi.test.junit5/src/main/java/org/osgi/test/junit5/properties/PropertiesConverter.java
```diff
@@ -0,0 +1,249 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *******************************************************************************/
+package org.osgi.test.junit5.properties;
+
+import static org.osgi.test.common.annotation.Property.Type.Scalar;
```
bjhargrave, line 20: Seems like this can be confused with the `import org.osgi.test.common.annotation.Property.Scalar;` below. Two things called `Scalar`. Perhaps we should not statically import this and just use `Type.Scalar`.

org.osgi.test.junit5/src/main/java/org/osgi/test/common/annotation/package-info.java
```diff
@@ -17,7 +17,7 @@
  *******************************************************************************/
 
 @Export(attribute = "junit=5", substitution = Substitution.NOIMPORT)
-@Version("1.1.1")
+@Version("1.2.0")
```
bjhargrave, line 20: Perhaps it would be good documentation to add `@since 1.2` javadoc comments to the new members/type introduced here?

org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/package-info.java
```diff
@@ -17,7 +17,7 @@
  *******************************************************************************/
 
 @Export(attribute = "junit=4")
-@Version("1.1.0")
+@Version("1.2.0")
```
bjhargrave, line 20: Perhaps it would be good documentation to add `@since 1.2` javadoc comments to the new members/type introduced here?

org.osgi.test.junit5/src/main/java/org/osgi/test/common/annotation/Property.java
```diff
@@ -55,6 +55,26 @@ public enum Type {
...
+		/**
+		 * Use the first {@link Property#value()} as a key in the Environment
+		 */
+		EnvironmentVariable,
+		/** Use the name of the test class */
+		TestClass,
+		/** Use the name of the test method */
+		TestMethod,
+		/** Use the test unique identifier */
+		TestUniqueId
```
bjhargrave, line 75: Is the unique id for a test fixed so that it can be known _a priori_? The test for this had to go through some contortions to "inject" the uniqueid into the test class. That is, is test unique id a practical source here that people would use? I am not sure.

bjhargrave (commented):
Yes, this looks very nice. I have a few comments.

org.osgi.test.junit5/src/main/java/org/osgi/test/common/annotation/Property.java
timothyjward: > Is the unique id for a test fixed so that it can be known a priori?

I believe that the answer to this is yes, but it's non-trivial to work out what it will be.

> That is, is test unique id a practical source here that people would use? I am not sure.

It was explicitly requested by @stbischof, but I do wonder if he actually wanted the Display Name. I'll wait for him to weigh in.

stbischof: The case for the TestUniqueId is `running parallel tests`.  I need a discriminator that i can use as ServiceProperty to mark a Service and use the same Id in a reference.target filter to bind the this service to an other.

When Running Parameterized in parallel they will have same TestClass and TestMethod.

org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/Property.java
laeubi: @bjhargrave if you look here at the original request:
- https://github.com/osgi/osgi-test/issues/614

this was explicitly mean that way that you specify a value that the can be overridden by a system property so this is actually intentional.

If one wants to go your suggestion, then we need some kind of evaluation here instead [like in maven](https://maven.apache.org/pom.html#Properties) that even allows default values (but no `Source`) or like [felix file install](https://felix.apache.org/documentation/subprojects/apache-felix-file-install.html#_property_substitution_in_configuration_files).

I'm just not sure this is a bit over engineered then, the idea was really that plain simple to allow to override the value if a system property of the given name exits, e.g. to specify a different default port for a webserver.

timothyjward: I think the change to use a `source` is a good idea, it allows us to support @stbischof in using test properties, and environment variables, without having to add an extra member each time.

The "default" is something that we've lost, but I think it can be easily (and elegantly) supported with a small change to make the second entry in the `value` array be the default if the system property doesn't exist. This would also apply to `EnvironmentVariable`

```
// Defaults to 8080 if web.port.prop is not set
@Property(key="foo", value= {"web.port.prop", "8080"}, source = SystemProperty, scalar = Integer)

// Fails if web.port.prop is not set
@Property(key="foo", value= "web.port.prop", source = SystemProperty, scalar = Integer)

```

laeubi: I just wanted to note that because this PR is marked to fix 
- https://github.com/osgi/osgi-test/issues/614

but nothing of that was requested nor being intended.

I'm not sure what @stbischof request is exactly about but maybe better solved / described in a different topic, I also don't really see how adding a new `SOURCE` type each time is better (what always is exclusive so one cant e.g. use env first, then system, then default) than an attribute ...

Also please note that for env variables there is already a notation of defaults, not sure if felix has something similar:
https://emmer.dev/blog/bash-environment-variable-defaults/

stbischof: @laeubi did you try the felix cm Plugin mechanism

https://github.com/apache/felix-dev/blob/master/configadmin-plugins%2Finterpolation%2FREADME.md

stbischof: I like the way you suggest tim.

// Defaults to 8080 if web.port.prop is not set
@Property(key="foo", value= {"web.port.prop", "8080"}, source = SystemProperty, scalar = Integer)

// Fails if web.port.prop is not set
@Property(key="foo", value= "web.port.prop", source = SystemProperty, scalar = Integer)

timothyjward: > I also don't really see how adding a new SOURCE type each time is better

The alternative would have been different members in the `Property` annotation to say `systemProperty` `environmentVariable`, `testClass`, `testName`...
This would explode the apparent complexity of the annotation, and lead to a lot of invalid setups (e.g. setting a `systemProperty` and a `testClass` at the same time).

Adding a `source` gives us one place where we define the rules for getting the value. I am in favour of avoiding extra plugins being needed for simple cases, but would draw the line at trying to replicate bash variable expansion.

org.osgi.test.junit5/src/main/java/org/osgi/test/common/annotation/Property.java
timothyjward: > The case for the TestUniqueId is running parallel tests. I need a discriminator that i can use as ServiceProperty to mark a Service and use the same Id in a reference.target filter to bind the this service to an other.

This says to me that we should leave the `TestUniqueId` in this PR. I think that the placeholder replacement should be raised as a separate issue so that it doesn't make this PR too big.

org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/Property.java
laeubi: Well its your to decide, but I don't see any need for testClass (why should it be for?) or env variables, but if it is desired, as said some kind of property expansion would be much more flexible, maybe reusing what @stbischof linked here:

https://github.com/apache/felix-dev/blob/master/configadmin-plugins%2Finterpolation%2FREADME.md

my request only included one simple thing, to override a value, it don't ask for env, testclass testname or any other things, If I would need such high level of felxibility and configuration I would simply not sue an annotation for that...

timothyjward:
I think that this PR is now ready.

Another issue should be created for handling template replacement as requested by @stbischof in https://github.com/osgi/osgi-test/pull/806#discussion_r1496664922

bjhargrave (approved):

PR #810, opened 2024-02-20
Fix slow testing with BlockingConfigurationHandlerImpl

### LINKED ISSUE #797: org.osgi.test.junit5.cm.test.ConfigAnnotationTest makes Tests lazy
I added a new TestClass to cm module.

When my Test is executed after `org.osgi.test.junit5.cm.test.ConfigAnnotationTest` it takes >60 sec.
Before it takes <10 sec.

I currently have no idea what this causes. Also not sure that problem is in `ConfigAnnotationTest`

timothyjward:
This fix increases test execution speed by avoiding the situation where the ConfigurationListener is unregistered from a CloseableBundleContext while the BlockingConfigurationHandler is still in the root context. Now registration is handled directly, and the BlockingConfigurationHandlerImpl is a CloseableResource.

Fixes #797


stbischof (approved):

stbischof:
great. thank a lot

PR #814, opened 2024-02-26
Support placeholders in configuration values

### LINKED ISSUE #812: Replacement in @WithFactoryConfigutation Properties and Factorys name
The case for the TestUniqueId is `running parallel tests`.  I need a discriminator that i can use as ServiceProperty to mark a Service and use the same Id in a reference.target filter to bind the this service to an other.

For marking this should work.
```
@Property(key = "testId", source = TestUniqueId)
```

But to define the target filter i would need a bit more. like an templatiting
```
@Property(key = "xyzService.target",value = "(testId=${value})", source = TestUniqueId),
```
or
```
@Property(key = "xyzService.target",value = "(testId=${xyz})", source = TestUniqueId, replacementKey=$"{xyz}"),
```
would also work with
```
@Property(key = "paramxyz",value = "NOT_SET", source = TestUniqueId, replacementKey="NOT_SET"),
```
When running parameterized in parallel they will have same TestClass and TestMethod.

This brings me to the Problem, that i also would need a generated name for my FactoryConfiguration because could not be the same on each test.
https://github.com/osgi/osgi-test/blob/32fbf7437d9451464e54782af019cd10c18300ef/org.osgi.test.junit5.cm/src/main/java/org/osgi/test/junit5/cm/ConfigurationExtension.java#L197-L204

timothyjward:
Sometimes it is not sufficient to use configuration values exactly as they exist in System Properties or other sources, or several properties need to be combined to create the configuration property. This commit adds support for Value Arguments to be included in the annotations. These arguments cause the Property value to be taken as a template string suitable for use with String.format, with the supplied arguments used to populate the template.

Fixes #812

### DIFF
--- a/org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/Property.java
+++ b/org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/Property.java
@@ -91,6 +91,21 @@ public @interface Property {
 		TestUniqueId
 	}
 
+	/**
+	 * Used to provide an argument in the {@link Property#templateArguments()}.
+	 * {@link ValueArgument} values are resolved in the same way as
+	 * {@link Property} values but they are always of {@link Type#Scalar}
+	 *
+	 * @since 1.2
+	 */
+	public @interface ValueArgument {
+		String[] value() default "";
+
+		Scalar scalar() default Scalar.String;
+
+		ValueSource source() default ValueSource.Value;
+	}
+
 	String key();
 
 	String[] value() default "";
@@ -104,5 +119,23 @@ public @interface Property {
 	 * @since 1.2
 	 */
 	ValueSource source() default ValueSource.Value;
+
+	/**
+	 * If any template arguments are set then the resolved value of this
+	 * {@link Property} annotation will be used as a template in
+	 * {@link String#format} with the resolved {@link ValueArgument} values used
+	 * as arguments.
+	 * <p>
+	 * Note that any defaulting or processing according to the {@link #source()}
+	 * will happen <em>before</em> the formatting is applied, so template
+	 * arguments cannot be used, for example, to change the name of a system
+	 * property. Conversely, the scalar conversion will happen <em>after</em>
+	 * the template is applied, meaning that numeric values can be assembled
+	 * from multiple template arguments.
+	 *
+	 * @return the arguments that should be used with the template
+	 * @since 1.2
+	 */
+	ValueArgument[] templateArguments() default {};
 
 }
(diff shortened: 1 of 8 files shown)


org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/Property.java
```diff
@@ -91,6 +91,21 @@ public enum ValueSource {
...
+	 * Used to provide an argument in the {@link Property#templateArguments()}.
+	 * {@link ValueArgument} values are resolved in the same way as
+	 * {@link Property} values but they are always of {@link Type#Scalar}
+	 *
+	 * @since 1.2
+	 */
+	public @interface ValueArgument {
+		String[] value() default "";
+
+		Scalar scalar() default Scalar.String;
```
bjhargrave, line 104: This does not seem to be used or tested.

org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/Property.java
```diff
@@ -105,4 +120,22 @@ public enum ValueSource {
...
+	 * will happen <em>before</em> the formatting is applied, so template
+	 * arguments cannot be used, for example, to change the name of a system
+	 * property. Conversely, the scalar conversion will happen <em>after</em>
+	 * the template is applied, meaning that numeric values can be assembled
+	 * from multiple template arguments.
+	 *
+	 * @return the arguments that should be used with the template
+	 * @since 1.2
+	 */
+	ValueArgument[] templateArguments() default {};
```
bjhargrave, line 139: There is a mismatch here between the element name (template) and the element value type (value).

org.osgi.test.junit4/src/main/java/org/osgi/test/junit4/properties/PropertiesConverter.java
```diff
@@ -41,60 +42,19 @@ private static Object toValue(Description desc, Property entry) {
...
-						Short shortVal = Short.valueOf(v);
-						val = primitive ? shortVal.shortValue() : shortVal;
-						break;
-
-					case String :
-						val = v;
-						break;
-				}
+			if (v != null && templateParams.length > 0) {
+				v = String.format(v, templateParams);
```
bjhargrave, line 55: The use of format strings opens up a mess of user errors (e.g. using non-String consuming conversions). Maybe it would be safer to use simple placeholders like logging. e.g. `{}`.

org.osgi.test.junit4/src/main/java/org/osgi/test/common/annotation/Property.java
timothyjward: I thought I had used it here:

https://github.com/osgi/osgi-test/pull/814/files#diff-14565fbe3b31cfc388c58178e1b351bc2a293b8eeaaf193a0af057125cfa9c42R363-R370

I'll fix up the test.

timothyjward:
A change for review comments, and also to cope with the wonder of inconsistent Java version numbering schemes!

bjhargrave (approved):
java version property test is still needing help :-)

stbischof (approved):
looks great! thx

Issue #816, opened 2024-02-28
Release 1.3.0


timothyjward:
There have been numerous fixes and enhancements since the release of 1.2.1, and the list of outstanding issues has been almost totally completed.

I propose that it's time to release 1.3.0 - what do you all think?

@stbischof @maho7791 @juergen-albert @bjhargrave @rotty3000 @kriegfrj

timothyjward:
Good to see that we have consensus - as for the process, is it necessary to create an RC1 tag first?

As far as I can tell the release process is as follows:

* Create two commits locally on main, the first changing the revision to a release version, the second to the next development version.
* Push the two commits together to main (to avoid a build of the release without a tag)
* Use the GitHub release tooling to tag the commit with the release version
* After the release has been built upload any artifacts to the GitHub release.

Are those steps correct?

maho7791:
Hi all,can you please wait until tomorrow. I wanted to test the issue with the empty configuration.On the first sight it only worked for Configurations but not for the FactoryConfiguration. I wanted to verify this.Regards,Mark Hoffmann M.A. Dipl.-Betriebswirt (FH) CEO/CTO Phone:   +49 3641 384 910 0 Mobile:  +49 175 701 2201 E-Mail: ***@***.*** Web: www.datainmotion.de Data In Motion Consulting GmbH Kahlaische Strasse 4 07745 Jena Germany Geschäftsführer/CEO Mark Hoffmann Jürgen Albert Jena HRB 513025 Steuernummer 162/107/05779 USt-Id DE310002614
-------- Ursprüngliche Nachricht --------Von: Tim Ward ***@***.***> Datum: 29.02.24  16:43  (GMT+01:00) An: osgi/osgi-test ***@***.***> Cc: Mark Hoffmann ***@***.***>, Mention ***@***.***> Betreff: Re: [osgi/osgi-test] Release 1.3.0 (Issue #816) 
Good to see that we have consensus - as for the process, is it necessary to create an RC1 tag first?
As far as I can tell the release process is as follows:

Create two commits locally on main, the first changing the revision to a release version, the second to the next development version.
Push the two commits together to main (to avoid a build of the release without a tag)
Use the GitHub release tooling to tag the commit with the release version
After the release has been built upload any artifacts to the GitHub release.

Are those steps correct?

bjhargrave:
> As far as I can tell the release process is as follows:

[...]

PR #830, opened 2024-03-13
Adjust compile versions for JUnit and AssertJ

timothyjward:
The compile versions of AssertJ and JUnit gave an inconsistent base version for JUnit Jupiter, and as AssertJ only optionally requires JUnit Jupiter (it really should be another module with a mandatory dependency 🙁) you can easily end up with broken dependency graphs at runtime. This commit changes the base for AssertJ back to 3.23.x and ups JUnit to 5.8.x, which is consistent. It means we lose a couple of JUnit versions, but 5.8.x has already been released for more than two years.

See #816 for more details

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,11 +37,11 @@
 		<maven.compiler.source>${java.version}</maven.compiler.source>
 		<maven.compiler.target>${java.version}</maven.compiler.target>
 		<bnd.version>6.4.0</bnd.version>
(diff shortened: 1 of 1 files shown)


stbischof:
LGTM

stbischof (approved):

bjhargrave (approved):
Make sure to include this dependency update in the release notes.

PR #1107, opened 2026-04-03
Add FrameworkQuiesce listener for waiting until framework events settle

stbischof:
Provides helper to block until no bundle or service events occur for a given period, simplifying stabilization in OSGi-based tests.

### DIFF
--- a/org.osgi.test.common/src/main/java/org/osgi/test/common/quiesce/FrameworkQuiesce.java
+++ b/org.osgi.test.common/src/main/java/org/osgi/test/common/quiesce/FrameworkQuiesce.java
@@ -1,0 +1,175 @@
+/*******************************************************************************
+ * Copyright (c) Contributors to the Eclipse Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *******************************************************************************/
+
+package org.osgi.test.common.quiesce;
+
+import java.time.Duration;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.ReentrantLock;
+
+import org.osgi.framework.AllServiceListener;
+import org.osgi.framework.BundleContext;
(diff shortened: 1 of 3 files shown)


stbischof:
@kriegfrj 
@timothyjward 

yould you please review

org.osgi.test.common/src/main/java/org/osgi/test/common/quiesce/FrameworkQuiesce.java
```diff
@@ -0,0 +1,175 @@
...
+
+	private void doWaitForQuiet(Duration quietPeriod, Duration timeout)
+		throws InterruptedException, QuiesceTimeoutException {
+		long quietNanos = quietPeriod.toNanos();
+		long deadlineNanos = System.nanoTime() + timeout.toNanos();
+		lock.lock();
+		try {
+			for (;;) {
+				long now = System.nanoTime();
+				if (now >= deadlineNanos) {
```
timothyjward, line 77: Is it your intent that we fail eagerly if the lock takes a while to acquire? Also it's better to compare short durations than clock points with nanoTime because it can overflow.

I think that I would restructure in the following way:

```java
if(quietPeriod.isNegative() || quietPeriod.isZero()) {
    throw new IllegalArgumentException("The quiet period duration must be positive");
}
if(timeout.compareTo(quietPeriod) <= 0) {
    throw new IllegalArgumentException("The timeout must be longer than the quiet period");
}
final long quietNanos = quietPeriod.toNanos();
final long deadlineNanos = timeout.toNanos();
final long start = System.nanoTime();
long elapsed = 0L;
long remainingQuiet = quietNanos;
lock.lock();
try {
    do {
        final long now = System.nanoTime();
        remainingQuiet = quietNanos - (now - lastEventNanos);
        if (remainingQuiet <= 0) {
            return;
        }
        elapsed = now - start;
    } while ((elapsed + remainingQuiet) <= deadlineNanos && 
            quiet.awaitNanos(remainingQuiet) > 0);
} finally {
    lock.unlock();
}
throw new QuiesceTimeoutException(quietPeriod, timeout, eventCount);
```

org.osgi.test.common/src/main/java/org/osgi/test/common/quiesce/FrameworkQuiesce.java
```diff
@@ -0,0 +1,175 @@
...
+	 *
+	 * @param bc the bundle context to register listeners on
+	 * @param quietPeriod required silence duration
+	 * @param timeout maximum total wait time
+	 * @return the number of events received during the wait
+	 * @throws InterruptedException if the waiting thread is interrupted
+	 * @throws QuiesceTimeoutException if {@code timeout} expires before the
+	 *             quiet period is reached
+	 */
+	public static int waitForQuiet(BundleContext bc, Duration quietPeriod, Duration timeout)
```
timothyjward, line 116: Should we avoid these static methods and make an injectable service?. That way we can encapsulate the `BundleContext` into the injected object rather than forcing the caller to inject a `BundleContext` just to wait for quiet.

